PR #39, opened 2014-03-03
Improved the way URLs are automatically detected in posts

pupi1985:
It seems parentheses bring a lot of trouble to actually perform an appropriate parsing. In short, I'm now removing the last closing parentheses if it doesn't have matching parentheses in the URL. These are the URLs I used to test:

(http://www.blahblah.com)
[http://www.blahblah.com]
http://www.blahblah.com
http://www.blahblah.com/sldkfjs/sdf.html
http://www.blahblah.com/sldkfjs/s&sdf%29df.html
http://www.blahblah.com. => This fails in 1.6.3
http://www.blahblah.com/.
http://wikipedia.org/wiki/sdf_sdf_(asd_asd_) => This fails in 1.6.3
http://blahblah.com/aha/aa752574(VS.85).aspx => This fails in 1.6.3

Related Q2A post: http://www.question2answer.org/qa/29015

### DIFF
--- a/qa-include/qa-app-format.php
+++ b/qa-include/qa-app-format.php
@@ -986,27 +986,40 @@
 	function qa_html_convert_urls($html, $newwindow = false)
 /*
 	Return $html with any URLs converted into links (with nofollow and in a new window if $newwindow).
-	Although they are valid URL components, the closing parentheses, if it is the last character of the URL is removed
-	from the URL. This is to avoid creating wrong broken in texts such as (http://www.question2answer.org) but will
-	allow URLs such as http://www.wikipedia.org/Computers_(Software) as the closing parentheses has a match in the URL.
+	Although they are valid URL components, the closing parentheses, curly and square brackets, if they are at the end
+	of the URL theyare removed from it if they don't have a matching opening one. This is to avoid creating wrong broken
+	URLs from texts such as (http://www.question2answer.org) but will allow URLs such as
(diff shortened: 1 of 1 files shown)


q2apro:
For this issue, is there any way (or the best way) to **prevent the parsing to a link**?

I have a special syntax for my users (javascript does some image magic), example:

`image(https://thesocietypages.org/socimages/files/2009/05/flickr-buddyicon.jpg) ...`

but it gets unfortunately parsed to:

`img(<a href="https://thesocietypages.org/socimages/files/2009/05/flickr-buddyicon.jpg">https://thesocietypages.org/socimages/files/2009/05/flickr-buddyicon.jpg</a>) ...`

Any way to prevent this?

q2apro:
Recent solution is to hack `format.php` and change:

```
function qa_html_convert_urls($html, $newwindow = false)
{
	// q2apro hack to not parse "bild(URL)" links - do hack
	$html = str_replace('image(https', 'image(htt ps', $html);
	
	// ...
	// q2apro hack to not parse "bild(URL)" links - undo hack
	$html = str_replace('bild(htt ps', 'bild(https', $html);

	return $html;
}
```	

This is doing the trick.

svivian:
You don’t need to hack the core, you can use a function override for that.

You could also use that override to actually do the conversion, instead of using JS.

Issue #158, opened 2014-12-02
Option to specify the separator for thousands and decimals


q2apro:
Fortunately, a member brought up the issue that I have problems with myself since using q2a: 

"How do we change the separator for thousands?"
http://www.question2answer.org/qa/41792/how-do-we-change-the-separator-for-thousands

It seems the layer/advanced theme cannot be used to change the output :-(

PR #188, opened 2015-01-10
[Improvement - LOW] Improvements in how numbers are displayed

pupi1985:
Continuing with the discussion in #158.

Ok, so these are the numbers I've modified. I guess those are all. Also note this doesn't include a feature to allow all numbers to be translated because that would mean all numbers should go through this function and there are many constants (particularly, a lot of number ones used in, e.g., "1_year") that I have not included. Anyway, it is a first step towards that direction.

A related item to mention is that I noticed this construction

```php
$value == 1
   ? qa_lang_html_sub('main/1_value', '1', '1')
   : qa_lang_html_sub('main/x_value', $value)
```

is repeated too often and I think it might be better to add some `qa_lang_html_sub_one_many($value, $langStringOne, $langStringMany)` function to reuse some code.

Regarding the "compacting" of numbers I've made a list of all the numbers I've sent through this function. I think compacting them would mean sending an additional parameter to that function but each place in which they're displayed need to be analyzed whether the compacting should happen or not. In the list, I've given my opinion based on when users would like the exact number and when they won't.

Modified numbers:
 * All numbers displayed in the favorite section (user points, tag amount, questions in category count): Compact
 * Alert displayed when exceeding file size number when attaching a big file in the ckeditor: Not compact
 * Votes in questions lists: Compact
 * Number of answers in question lists: Compact
 * Number of views in questions lists and question pages: Compact
 * Amount of points displayed in the <who> section for question info. EG: "ASKED Nov 26 in Cat 1 by admin (123,716 points)": Compact
 * Amount of points displayed in the <who> section for answer, comment or edit info. EG: "ANSWERED Nov 26 in Cat 1 by admin (123,716 points)": Compact
 * Number of questions per category in the (fake) category widget: Compact
 * Number of emails sent when sending mass emails (the updated email numbers): Not compact
 * In Admin -> Categories 
   * When editing a category, the total question count: Not compact
   * When editing a category, the amount of questions for a category that have no subcategory: Not compact
   * When there are questions with no category and the setting that requires question to have a category is enabled: Not compact
 * In Admin -> Stats:
   * Database size: Not compact
   * Total questions: Not compact
   * From users: Not compact
   * From anonymouns: Not compact
   * Total answers: Not compact
   * From users: Not compact
   * From anonymouns: Not compact
   * Total comments: Not compact
   * From users: Not compact
   * From anonymouns: Not compact
   * Registered users (for internal users): Not compact
   * Active users (for external users): Not compact
   * Users who posted: Not compact
   * Users who voted: Not compact
   * When reindexing content (the updated pages, posts and words reindexed): Not compact
   * When recounting content (the votes and flags and answers and hotness recount): Not compact
   * When recalculating user points: Not compact
   * When refilling event streams: Not compact
   * When recalculating categories (): Not compact
   * When deleting hidden posts (questions, answers and comments): Not compact
   * When moving blob to disk or db: Not compact
 * In Admin -> Users when displaying points of the users' titles: Not compact
 * In categories page, the amount of questions per category: Compact
 * In tags page, the amount of questions per tag: Compact
 * In users page, the amount of points per user: Compact
 * In profile section
   * Score: Not compact
   * Questions: Not compact
   * Answers: Not compact
   * Comments: Not compact
   * Voted on questions and answers: Not compact
   * Gave out up and down votes: Not compact
   * Received up and down votes: Not compact
   * Rank: Not compact
   * Question with best answer selected: Not compact
   * Answers selected as best: Not compact
 * Activity count widget: Compact
 * Maximum file size in wysiwyg plugin options: Not compact

Regarding the setting to control whether to compact numbers or not, should it be placed at the bottom of the admin/viewing section or in any other place?

Slightly unrelated to this branch, I've also mixed a refactor of `qa_recalc_get_message` that adds a "private" function. I think it is a good way to avoid issues in global context for functions that should not be part of the core's interface. More info in this comment: https://github.com/q2a/question2answer/pull/101#issuecomment-61041024

BTW, the easy way to test this would be to multiply the number by 1000 and change the thousand and decimal keys in the language file.

Let's keep this PR open (unmerged) until it is ready so that I can rebase it freely, if needed. Thanks.

### DIFF
--- a/qa-include/app/favorites.php
+++ b/qa-include/app/favorites.php
@@ -123,6 +123,7 @@
 			return null;
 
 		require_once QA_INCLUDE_DIR.'app/users.php';
+		require_once QA_INCLUDE_DIR.'app/format.php';
 
 		$ranking = array(
 			'items' => array(),
@@ -137,7 +138,7 @@
 			$ranking['items'][] = array(
 				'avatar' => $avatarhtml,
 				'label' => $usershtml[$user['userid']],
-				'score' => qa_html(number_format($user['points'])),
+				'score' => qa_html(qa_format_number($user['points'])),
 				'raw' => $user,
 			);
 		}
@@ -148,9 +149,11 @@
 
 	function qa_favorite_tags_view($tags)
 /*
-	Returns content to set in $qa_content['ranking_tagss'] for a user's favorite $tags.
+	Returns content to set in $qa_content['ranking_tags'] for a user's favorite $tags.
 */
 	{
+		require_once QA_INCLUDE_DIR.'app/format.php';
+
 		$ranking = array(
 			'items' => array(),
 			'rows' => ceil(count($tags)/qa_opt('columns_tags')),
@@ -160,7 +163,7 @@
 		foreach ($tags as $tag) {
 			$ranking['items'][] = array(
 				'label' => qa_tag_html($tag['word'], false, true),
-				'count' => number_format($tag['tagcount']),
+				'count' => qa_html(qa_format_number($tag['tagcount'])),
 			);
 		}
 
@@ -173,6 +176,8 @@
 	Returns content to set in $qa_content['nav_list_categories'] for a user's favorite $categories.
 */
 	{
+		require_once QA_INCLUDE_DIR.'app/format.php';
+
 		$nav_list_categories = array(
 			'nav' => array(),
 			'type' => 'browse-cat',
@@ -182,7 +187,7 @@
 			$cat_url = qa_path_html( 'questions/' . implode( '/', array_reverse(explode('/', $category['backpath'])) ) );
 			$cat_anchor = $category['qcount'] == 1
 				? qa_lang_html_sub('main/1_question', '1', '1')
-				: qa_lang_html_sub('main/x_questions', number_format($category['qcount']));
+				: qa_lang_html_sub('main/x_questions', qa_format_number($category['qcount']));
 			$cat_descr = strlen($category['content']) ? qa_html(' - '.$category['content']) : '';
 
 			$nav_list_categories['nav'][$category['categoryid']] = array(
--- a/qa-include/app/favorites.php
+++ b/qa-include/app/favorites.php
@@ -138,7 +138,7 @@
 			$ranking['items'][] = array(
 				'avatar' => $avatarhtml,
 				'label' => $usershtml[$user['userid']],
-				'score' => qa_html(qa_format_number($user['points'])),
+				'score' => qa_html(qa_format_number($user['points'], 1, true)),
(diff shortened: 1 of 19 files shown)


svivian:
From a quick look this is all excellent! Only thing that sticks out to me is the length of the language keys (maybe you were following `date_format_other_years` style?). Perhaps just `num_decimal_point` and `num_thousands_separator` would suffice? Or starting with an underscore like `_decimal_point` to indicate it's a special translation and not the string "decimal point".

Interesting point about the 'private' functions. What you've got seems to work fine. There is also a tag for docblock comments, `@access private` that we could use, either instead of the function name addition, or as well as it.

pupi1985:
I've changed the texts of the language strings (I have to confess I like long an overdescriptive strings) and I've added the ability to compact numbers. The ones compacted are the ones specified in the list above. Let me know if I'm missing anything. Thanks.

svivian:
Hey @pupi1985 I'm just working on merging this at the moment. I think there may be a problem with the decimal places. When I call `qa_format_number(5.452, 1)` it returns 5 rather than the expected 5.5.

The problem stems from the `$decimals` parameter being used for compact numbers as well as regular numbers. We pass in 1 decimal so we end up with "9.1k" for example. But if the number doesn't get compacted, it's set to 0 decimals, which messes with regular numbers.

On the flipside, removing that else clause means that `qa_format_number(9123, 1, true)` will return "9,123.0" if it doesn't get compacted.

Can't think of a way round this at the moment...

svivian:
OK so I wrote a bunch of tests for expected behaviour, and refactored the function to make them work. (Unit tests have never been so useful!)

If numbers are compacted then 1 decimal place is used (a .0 is still removed) and the `$decimals` parameter is ignored.

If numbers are not compacted then `$decimals` takes effect. Which means this...

> On the flipside, removing that else clause means that qa_format_number(9123, 1, true) will return "9,123.0" if it doesn't get compacted.

...is actually correct behaviour, but the functions in Q2A shouldn't be called like that. If the number is an int and could be compacted, we should call it with `$decimals=0`.

pupi1985:
That's great Scott. Sorry for the delay but these days I felt like 24 hours a day just didn't seem to be enough!

Indeed tests were the key here to better understand requirements themselves. Anyway, the results seem pretty good to me and that's what it actually matters in the end :) We'll have to wait for 1.8 testers to hear their feedback about it, though.

I don't want to minimize this whole work but believe me or not the most important thing here, IMO, is `@access private`. That basically is a warning for plugin developers that tells them to use that function at their own risk: it might change in any way without notice. That really simplifies things when needing to change function declarations (e.g.: changing parameters types, or even adding or removing them).

PR #193, opened 2015-02-01
[Improvement - MED] Added a theme-dependant CSS class to the body tag

pupi1985:
The motivation of the PR is to allow plugin developers to distribute plugins that work out of the box.

Currently, each user has to configure the appropriate CSS for each theme. I decided to provide the CSS for each (popular) theme but each user has to apply it. Alternatively, each plugin developer could check the currently selected theme (by means of PHP) and select a given CSS to use. I think a better approach would be to fix a presentation issue just with CSS.

This PR sets a theme identifier (which also needs to be a valid CSS classname) to the body of the HTML document. That way each plugin developer can provide a CSS file with the general stuff (theme independent CSS) and the fixes for each (popular) theme (theme dependent CSS).

A default value should be applied in case a theme does not have an identifier configured. EG:
```
.qa-theme-snowflat .my-button {
    padding: 20px;
}

.qa-theme-default .my-button,
.qa-theme-classic .my-button {
    padding: 0px;
    margin: 10px;
}
```
The good thing is that this change is 100% backwards compatible and, speaking for myself, I would use this in each plugin that includes CSS.

### DIFF
--- a/qa-include/qa-theme-base.php
+++ b/qa-include/qa-theme-base.php
@@ -48,7 +48,7 @@ class qa_html_theme_base
 	protected $rooturl;
 	protected $request;
 	protected $isRTL; // (boolean) whether text direction is Right-To-Left
-	protected $theme = 'default';
+	protected $theme;
 
 	// whether to use new block layout in rankings (true) or fall back to tables (false)
 	protected $ranking_block_layout = false;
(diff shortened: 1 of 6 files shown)


svivian:
Interesting idea. But I wonder if doing it like this is going to bloat people's CSS by adding styles for 4+ different themes?

There is also an issue with your CSS - what if the user has a custom theme with their own identifier? I think you'd want

    .my-button {
        padding: 0px;
        margin: 10px;
    }
    .qa-theme-snowflat .my-button {
        padding: 20px;
    }

Which goes some way to solving the first issue as you only have overrides for certain themes. Most of the CSS stays the same.

pupi1985:
You're right, I totally missed that. So it wouldn't even make any sense to have a `default` theme setting as the CSS selector without a theme would cover both: themes with identifiers not taken into account by the plugin developer and themes without a plugin identifier.

svivian:
Implemented this separately due to conflicts. Thanks for the PR.

PR #194, opened 2015-02-03
[Improvement - LOW] Minor refactorings to the SnowFlat theme

pupi1985:


svivian:
Good PR, I was planning on removing that extra PHP file at some point :)

pupi1985:
Thanks. The only thing I like more than adding code... is removing it :)

PR #197, opened 2015-02-18
Patch to core login to allow login credential check to be overridden in a plugin

sweetlilmre:
In this patch, the core credential check functionality has been moved into a function.
This allows a plugin e.g. the ldap plugin (which I have also patched appropriately under my fork) to very simply override the credential check and login against LDAP whilst utilizing the existing login handling framework.


sweetlilmre:
Sorry, wrong branch will submit to dev

PR #198, opened 2015-02-18
[PATCH] Moved core credential check into function in app/users.php

sweetlilmre:
The intention of this patch is to refactor the core login credential check into a function.
This in turn allows a plugin to override the check and provide different / extended functionality.
This is particularly useful in the case of the LDAP plugin (to which I am a contributor).
Previously the plugin could only be installed via a modification to a core page. Now the plugin can simply override the credential check function and hook the LDAP credential check process into the core login framework seamlessly.

### DIFF
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -225,7 +225,7 @@
 			unset($_SESSION['qa_session_verify_'.$suffix]);
 		}
 
-		function qa_check_login_credentials($inemailhandle, $inpassword)
+		function qa_check_login_credentials($inemailhandle, $inpassword, $inremember)
 	/*
 		Call to check login credentials, override in plugin for custom credential checking whilst maintaining the normal cookie handling etc.
 		returns:
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -51,6 +51,9 @@
 	@define('QA_FORM_EXPIRY_SECS', 86400); // how many seconds a form is valid for submission
 	@define('QA_FORM_KEY_LENGTH', 32);
 
+	const LOGIN_CREDENTIALS_OK = 0;
+	const LOGIN_USER_NOT_FOUND = 1;
+	const LOGIN_PASSWORD_INCORRECT = 2;
 
 	if (QA_FINAL_EXTERNAL_USERS) {
 
@@ -229,9 +232,9 @@
 	/*
 		Call to check login credentials, override in plugin for custom credential checking whilst maintaining the normal cookie handling etc.
 		returns:
-		0 == credentials OK
-		1 == user not found
-		2 == password incorrect
+		LOGIN_CREDENTIALS_OK (0) == credentials OK
+		LOGIN_USER_NOT_FOUND (1) == user not found
+		LOGIN_PASSWORD_INCORRECT (2) == password incorrect
 	*/
 		{
 			if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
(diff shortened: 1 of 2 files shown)


sweetlilmre:
Please see: https://github.com/sweetlilmre/qa-ldap-login
for the ldap plugin and how this patch is utilized.

svivian:
Hi, thanks for the contribution! I don't really know anything about LDAP so I have to ask, is this something that really cannot be achieved with the existing Q2A plugin system?

Q2A has both [login plugins](http://www.question2answer.org/modules.php?module=login) as well as a [single sign-on system](http://www.question2answer.org/single-sign-on.php) that should be able to handle your use case without core hacks. Is there something specific that they cannot do?

sweetlilmre:
Hi, thanks for the quick response!

I don't believe that the existing functionality can quite deliver, here is the rationale:

LDAP is an odd one. The single sign-on systems like FB, or forums etc. have an existing DB against which a user can register / login with associated URLs etc. LDAP doesn't fit into this model as user management is handled at the domain level. Admittedly it might be possible to create a shim / proxy to present to Q2A that would partially behave in this manner, but this would seem inelegant.

One of the primary requirements for an LDAP integrated system is that ONLY LDAP credentials may be used to create accounts and authenticate the users (this only exception being fallback to exisitng base accounts for administration purposes, which the LDAP plugin supports). The intention here is to enable that use-case without having to hack the Q2A system and hence the refactor to allow a plugin override.

The login plugin approach is almost there (the original LDAP plugin partially went this route), but the focus (as far as I can understand it) is to provide your own separate login / cookie etc. management and hence the page hack that was required to insert LDAP into the system as the only authentication mechanism.

Additionally I think that this patch could simplify the integration of other login systems. The LDAP patch is has been significantly simplified and the simplification seems to be generally applicable.

I am hoping that as the patch is a pure refactor and doesn't change behaviour that it will be accepted :)

sweetlilmre:
Hi,

Have you had any further thoughts on this?

pupi1985:
I took a quick look this, and I still can't fully figure out why wouldn't external users fit in this situation. Understanding that is the key here.

From a birds eye view, it seems you just want to reuse the login page to perform your custom login. If that is the case maybe you should replace the login page with your own one. EG: What about the forgot password? Which of the 2 passwords should the user receive?

Now, assuming for some reason they don't. Then you should be aware that you will be "duplicating" users as you will have the ones in your LDAP and the ones in Q2A as well. That could complicate things when users change their username, password or email in Q2A and/or the LDAP and expect they would be replicated in the other system.

Despite this particular case I think there should be some refactors there oriented towards what you suggest. In fact, this could open a different discussion about how to start using exceptions, as this seems to be the ideal case to use them.

sweetlilmre:
@pupi1985 thanks for your comments.

In my case, there shouldn't be any duplicate users as normal registration is disabled via another override (in the LDAP plugin). All users are mapped directly to LDAP and so password changes are also not an issue.

I will re-look at the external user mechanism, it didn't seem to work when I first looked into it, but perhaps with a fresh view I can find a way. The ideal situation would be something along the lines of:
  all login / registration mechanisms are considered 'external'
Then there wouldn't be exceptions per se :)

svivian:
After thinking about this for a while (sorry I left it so long), I don't think it makes sense to allow overriding the core Q2A login.

I'm sure you can implement this using Single Sign On. You can add a page plugin with a login form easily, and add its URL to the `qa_get_login_links` function in `qa-external-users.php`. That same plugin can handle the actual authentication that you wanted to add to the function override from this PR.

Thanks for the contribution.

PR #208, opened 2015-04-09
[Bugfix - MED] Unable to recategorize closed questions

pupi1985:
If a question is closed and there is a limit in the amount of characters a question can have in the title or in the content then it can't be recategorized.

The commit to blame is this one: 3891fdd8fd72542c8247fa7b58a98f37dfd13d0e

Now, considering different validation rules apply to questions that are closed and to questions that are open, I thought on splitting the logic and the validation of fields. This results in the closed question not having their title and content fields validated.

Note the [filter module](http://www.question2answer.org/modules.php?module=filter) documentation should be updated.

### DIFF
--- a/qa-include/pages/question-post.php
+++ b/qa-include/pages/question-post.php
@@ -458,13 +458,19 @@
 		else {
 			$in['queued']=qa_opt('moderate_edited_again') && qa_user_moderation_reason($userlevel);
 
-			$filtermodules=qa_load_modules_with('filter', 'filter_question');
-			foreach ($filtermodules as $filtermodule) {
-				$oldin=$in;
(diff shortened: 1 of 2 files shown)


svivian:
I think this is the wrong solution. In a filter plugin, the `$question` variable is supposed to only contain the fields being edited. In other words, if `$question['title']` is null then we should be ignoring it (as was the case previously). However, this gives the previous bug where not submitting a title at all would be allowed through.

So we need different behaviour when a question is being edited or added for the first time. I think I have a solution which I will work on tomorrow.

svivian:
OK this should be fixed in the commits from ea166e4b32205e32fbeead3c77c4a7af66bcd786 onwards. Let me know if you spot any problems :)

PR #210, opened 2015-04-13
[Improvements - LOW]  Some minor refactorings

pupi1985:
IMO this [line](https://github.com/q2a/question2answer/compare/q2a:dev...pupi1985:patch-51?expand=1#diff-d038fdbb231a140f35fccebed753de50R918) should be:
```php
$error = 'login';
```

If a user is not logged in, it makes no sense to display a 'level' error.

PS: I did keep the logic this time, according to some brute force tests I've made.

### DIFF
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -873,47 +873,50 @@
 			$error=isset($userid) ? false : 'login';
 
 		elseif ($permit>=QA_PERMIT_CONFIRMED) {
-			if (!isset($userid))
+			if (isset($userid)) {
+				if (
+					QA_FINAL_EXTERNAL_USERS || // not currently supported by single sign-on integration
+					($userlevel>=QA_PERMIT_APPROVED) || // if user approved or assigned to a higher level, no need
+					($userflags & QA_USER_FLAGS_EMAIL_CONFIRMED) || // actual confirmation
+					(!qa_opt('confirm_user_emails')) // if this option off, we can't ask it of the user
+				)
+					$error=false;
+				else
+					$error='confirm';
+			} else
 				$error='login';
-
-			elseif (
-				QA_FINAL_EXTERNAL_USERS || // not currently supported by single sign-on integration
-				($userlevel>=QA_PERMIT_APPROVED) || // if user approved or assigned to a higher level, no need
-				($userflags & QA_USER_FLAGS_EMAIL_CONFIRMED) || // actual confirmation
-				(!qa_opt('confirm_user_emails')) // if this option off, we can't ask it of the user
-			)
-				$error=false;
-
-			else
-				$error='confirm';
 
 		} elseif ($permit>=QA_PERMIT_APPROVED) {
-			if (!isset($userid))
+			if (isset($userid)) {
+				if (
+					($userlevel>=QA_USER_LEVEL_APPROVED) || // user has been approved
+					(!qa_opt('moderate_users')) // if this option off, we can't ask it of the user
+				)
+					$error=false;
+				else
+					$error='approve';
+			} else
 				$error='login';
 
-			elseif (
-				($userlevel>=QA_USER_LEVEL_APPROVED) || // user has been approved
-				(!qa_opt('moderate_users')) // if this option off, we can't ask it of the user
-			)
-				$error=false;
+		} else {
+			if (isset($userid)) {
+				if ($permit>=QA_PERMIT_EXPERTS)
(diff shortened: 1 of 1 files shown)


svivian:
If I'm honest, I'm not sure if you've improved things here (though you haven't made them worse). Both old and new have a lot of nested logic.

I wonder if it would be much simpler to use an "early return" system. So instead of setting `$error` then returning it at the bottom, we return the value right away. Ignoring the final if clause, `$error` can only be set once.

P.S. what are your "brute force" tests? Would they fit into our test suite? This function just takes some numbers and outputs a string so it should be a perfect candidate for a unit test.

pupi1985:
I've only made very minor changes indeed. Only some performance improvements by removing repeated checks of the `$userid` variable. Anyway, the main reason for these changes was actually to expose what I've commented above: I don't think the function should return 'level' if the user is not even logged in.

Similarly, this one I haven't actually suggested it in the code, the `'userblock'`, IMO, shouldn't care at all about `'level'`. If the user is logged in and is blocked then it should be denied because they are a blocked rather than because they are not approved or require their email confirmation. Removing that condition would allow that `if` to be moved to the top and perform an early return. But that would have implied a change in the logic.

Regarding the tests I think I have them somewhere (I think I've stashed them at home). However, I really mean it when I say a brute forced them. There are MANY combinations of data per parameter: may permits, 2 different states for user ids (null or assigned), many user levels and 2 user flags being used which means there are at least 4 different combinations of flags to consider. You should also take into account that there are 2 settings from the `^options` table and the external users check that, as far as I understand PHP, there is nothing to do to test that in its 2 possible states (I've mentioned this in another PR).

So there are quite a lot of tests and trying to figure out which ones to remove was harder than brute force all the possible values derived from the previously mentioned combinations. They were over 2K different combinations. I'll see if I have the code somewhere, at least the one to generate the tests and then you decide :)

Alternatively, maybe this function could also be split into smaller, almost meaningless functions, that could be tested independently. But that's your call :)

pupi1985:
Found it: https://gist.github.com/pupi1985/8c7870e90567c3b308be

It is not elegant but got the job done without much thinking :)

svivian:
I've just committed some basic tests with the current behaviour.

The 'level' error, I can see it both ways. Since Q2A can be set up to allow anonymous posting, not being logged in is a level status. Returning an error saying they need to be logged in is potentially misleading because they need a higher permission, not just to log in.

But the message "You do not have permission to perform this operation" isn't much clearer. And it is inconsistent since general user operations return 'login' even when they need to be approved users or have their email confirmed.

The other anomaly I noticed doing the tests is that if a user is blocked it returns 'userblock' even when the permission is 'QA_PERMIT_ALL'. Which is silly because it means they can just log out and do that operation! That should probably be fixed.

pupi1985:
I can tell you got what I meant because you used the word *inconsistent*. Anyway, I hadn't thought about the fact that anonymous users can still perform operations. Maybe this all depends on what error should the user receive in case they can't perform a certain operation: the highest one (e.g.: level) or the lowest one (e.g.: login). IMO, only logged in users should receive a 'level' and, for anonymous users, I think it is perfectly fine for the user to be asked to be logged in to perform an operation. Even after doing so, I think it is fine to tell them they don't have the appropriate permissions. I'd be fine with any approach but it should be consistent regardless of the requested permit.

Regarding blocked users logging out and performing anonymous operations... I don't really thing there is a way to fix that. In fact, that's how "requirements" are "defined". Any user (person) can be an anonymous user. Even people who have a blocked user. So it is OK that anonymous users could post questions. I even think it is useful to block users in a site that allows anonymous posting because you might not care whether those people (not users!) post or not but you do care about them voting, for instance.

Taking it a step further, in technical terms, it is not possible to block the person, only the user. You can make it harder for the person, e.g., validating cookies, but not impossible.

svivian:
Thanks for your work on this, it's now been merged.

I changed that 'level' error to a 'login' one. If a logged out user visits a page for Editors/Admins etc then it makes more sense to show the "please login" prompt in case the user is actually an Editor/Admin. After making this change a lot of the logic in the function could be simplified (thanks to your brute force script I could verify the results were the same besides said level/login change).

As for the blocked user/anonymous scenario, I think it does make sense like that after all. Even if you allow anonymous posts you may still want to prevent a user making their posts with their account. Plus if you have a CAPTCHA for anons, banning a spam account forces them to prove they're human.

pupi1985:
Yeah, the good brute force script. I never leave home without one of those.

Definitely, the function seems easier to read read now :)

PR #211, opened 2015-04-13
[Improvements - MED] Approach to remove global variables and to enable safe mode

pupi1985:
Just throwing some ideas about configuration management and cache storage.

I thought this could be a feasible approach to get rid of global variables as well as providing a way to change configurations to execute test cases. Currently, it is not possible to create a test a of a function that queries a constant (e.g.: checking for external users) as the constant can't be changed. Having the ability to change those constants allows tests to be created. Or maybe there is another way to workaround this that I couldn't figure out.

Note that although the "global" keyword is gone, a "global" context is still used.

Regarding safe mode, it is just an excuse to see how this approach works, but probably might be useful anyway.

PS: To better see the diff in admin-plugins.php it's better to perform a `git diff -w`

### DIFF
--- a/qa-include/Q2A/Application.php
+++ b/qa-include/Q2A/Application.php
@@ -5,7 +5,7 @@
   http://www.question2answer.org/
 
   File: qa-include/Q2A/Application.php
-  Description: Contains configuration data of Q2A
+  Description: Contains configuration and cache data of Q2A
 
 
   This program is free software; you can redistribute it and/or
@@ -24,6 +24,7 @@
 class Q2A_Application {
 
 	private $config;
+	private $cache;
 	private static $instance;
 
 	/**
@@ -42,6 +43,7 @@ class Q2A_Application {
 	 */
 	private function __construct() {
 		$this->config = new Q2A_Util_Set();
+		$this->cache = new Q2A_Util_Set();
 
 		$this->config->set('safe_mode', false);
 	}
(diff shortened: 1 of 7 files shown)


svivian:
Great idea, I'd also been thinking about how to get rid of globals. The safe mode is useful but I don't see a way to turn it on/off, what was the use case you were thinking of? An option in admin, or a flag set by the developer in the config file (like QA_DEBUG_PERFORMANCE)?

Only very minor thing is getCache/setCache seems a little too "implementation specific". In other words, it wouldn't matter to the caller whether it's cached or not, they just want to get/set data. Though I haven't thought of a better name as yet...

pupi1985:
Regarding safe mode, actually the I thought there shouldn't be a way of turning it off. I mean, If you require a safe mode, something has probably messed seriously with the installation. Maybe a theme or plugin breaking the admin page. So if you rely on the UI to turn safe mode on then it is not so safe :) I don't think setting it in the constructor (or somewhere to initialize it) is a bad a idea.

Relating it to QA_DEBUG_PERFORMANCE, if it was up to me, I would remove all those constants and would add them as an initialization of the Q2A_Application class. That would allow configuration to be mocked.

Regarding the cache, I don't care much :P Anyway, just wanted to make it clear that data in there is not reliable at all times. I'm also thinking whether some elements should be unset/disposed or not. E.G.: those layers after being processed by the theme and the hierarchy created and processed etc, they don't seem to be of much use.

Anyway, again, my intention here was just to start thinking on this. I know the PHP4 style is all over the core and thinking in turning into PHP5, using OOP, encapsulation, data hiding (which turned out to be a significant issue since 1.7, based on other developer's comments), etc., is a such a long way that sometimes it is hard to see where to start. This is just a step on that direction :)

svivian:
Gonna (finally) close this now since it's mostly been implemented.

PR #218, opened 2015-04-20
[Bugfix - LOW] Removed the Search text from the input search button

pupi1985:
I don't really like this approach but as long as there is a blank character between those 2 inputs any approach would be a mess. With this one, at least, you forget about the width of the blank character between the inputs.


svivian:
Fixed this a slightly different way, by inverting the text indent (should have been negative in the first place).

PR #232, opened 2015-05-11
Update README.md

gurjyot:
This will help in getting way many users for Q2A on github by listing all the features of Q2A.


svivian:
Thanks for the great idea. I decided to add the feature list from the Q2A home page as it was a little more concise.

gurjyot:
I am happy that you liked the idea. And yes your features list is very much complete and detailed so it's better to have that... :-)

PR #235, opened 2015-05-16
[Bugfix - LOW] Corrected a bug that bypassed checking the allow_multi_answers option when posting

pupi1985:
How to reproduce the issue:

1.  Disable the `allow_multi_answers` setting
1.  Open a question that has no answer for your current user in 2 separate tabs
1.  Post an answer in one tab
1.  Post an answer in the other tab

Expected result: The answer is not posted
Actual result: :boom: 

Although this issue is actually related to the model rather than the view, the current design is focused on the view and whether visual components, such as buttons, should be displayed or not. So in order not to reinvent the will for a simple bugfix I think the best approach would be to take that business logic from the view handling functions.

### DIFF
--- a/qa-include/pages/question-submit.php
+++ b/qa-include/pages/question-submit.php
@@ -353,6 +353,16 @@
 			}
 
 			if (empty($errors)) {
+				// Makes sure users don't add more than one answer when allow_multi_answers is off
+				require_once QA_INCLUDE_DIR . 'pages/question-view.php';
+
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for the PR. I thought that calling the monolithic `qa_page_q_post_rules` function was a bit much - we only needed a tiny bit of logic from there - so I've done this a slightly different way. See e43e353, let me know if you find any problems.

PR #237, opened 2015-05-19
update qa-style.css

audacioustux:
for avoiding horizontal scroll at 1035px


svivian:
Thanks for the PR, I fixed this in a slightly different way.

audacioustux:
oh... greate :smile:

PR #240, opened 2015-05-22
ip related changes

toby78:
- i have added support for proxies, cloudflare cdn
- changed all ips to ipv6 and converted ipv4 to ipv6 so it works both in an ipv4 and ipv6 world

### DIFF
--- a/qa-external-example/qa-external-users.php
+++ b/qa-external-example/qa-external-users.php
@@ -20,6 +20,7 @@
 	More about this license: http://www.question2answer.org/license.php
 */
 
+require_once QA_INCLUDE_DIR.'db/users.php';
 
 /*
 	=========================================================================
@@ -51,7 +52,7 @@
 
 	//	Set this before anything else
 
-		return null;
+		return "INT";
 
 	/*
 		Example 1 - suitable if:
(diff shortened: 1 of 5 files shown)


svivian:
Hi toby78, thanks for the effort you've put in. Unfortunately I cannot accept this PR in its current state for a few reasons.

Firstly, this contains several changes specific to you e.g. the external users implementation, which of course should not be in Q2A itself. So you'll need to remove those commits.

I don't think it makes sense to support CloudFlare in core Q2A. CF is just one CDN and we can't support every individual setup in the core. The IP function `qa_remote_ip_address` is extendable so you can create a function override for your specific case. See [the docs](http://www.question2answer.org/overrides.php) for details.

Along similar lines, it's not appropriate to use `$_SERVER['HTTP_X_FORWARDED_FOR']` because that is just an HTTP header and cannot be trusted. If you have your own proxy server you'd need to first check that `REMOTE_ADDR` is that of your proxy and then check `HTTP_X_FORWARDED_FOR`. Again, you can do this in a function override; it's not applicable to most people.

I'm interested in supporting IPv6 so if you're able to extract those commits from your pull request I'll take a look at merging them. Thanks.

PR #243, opened 2015-05-23
[Theme] Removed the link styling from the categories in the level user profile field

pupi1985:
The `#level` field in the user profile takes the styling of the private message link to all the links inside, including the category-specific ones:

download

Alternatively, the whole "button styling" could be removed and keep them all as links.
More alternatively, the way of contacting a user should not be in the level field. It seems it ended up there just because there was some space available.

### DIFF
--- a/qa-theme/SnowFlat/qa-styles.css
+++ b/qa-theme/SnowFlat/qa-styles.css
@@ -2036,7 +2036,7 @@ input[type="submit"], button {
 	margin-top: 0;
 }
 
-#level .qa-form-wide-static a {
+#level .qa-form-wide-static a:first-child {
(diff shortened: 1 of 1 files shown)


gurjyot:
Can something be done without removing styling for message button. It actually catches the attention of a user towards private message feature. Secondly having a button for message looks good instead of just a boring link.

svivian:
I agree it doesn't make sense for the "private message" button to be there. What is the other button in your screenshot, is that from a plugin?

pupi1985:
The other button is actually a permit per category, which is part of the core.

I agree on not styling by IDs. I just proposed the change with the minimum impact, trying to avoid modifying PHP code.

PR #250, opened 2015-06-06
[Refactor] Minor changes to the mouse-layer plugin

### LINKED ISSUE #248: Issue of mouseover-layer plugin
Refer to thread below.
http://www.question2answer.org/qa/45666

pupi1985:
Minor refactor of `qa_shorten_string_line`
Minor changes to the mouse-layer plugin
  1. Fix #248 (text being cut twice)
  1. Split presentation logic
  1. Removed duplicated blanks, tabs, new lines, etc to compact the string

### DIFF
--- a/qa-include/util/string.php
+++ b/qa-include/util/string.php
@@ -495,14 +495,14 @@
 	 * @param string $ellipsis Text used to replace the removed words from the original text
 	 * @return string The string turned into a single line and cut to fit the given length
 	 */
-	function qa_shorten_string_line($string, $length)
+	function qa_shorten_string_line($string, $length, $ellipsis = ' ... ')
 	{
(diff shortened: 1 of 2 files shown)


svivian:
I'm not following what the problem is. The text is cut in the SELECT query so that we don't fetch copious amounts of text over MySQL then just throw it away. Is there a need to select everything?

pupi1985:
If you see how the ellipsis are placed in function `qa_shorten_string_line` you'll see that you actually need the whole text...
Alternatively, the ellipsis could be placed at the end and get rid of all that complex logic :)

svivian:
OK guess I was having some brain fade there. I think it does make more sense to get the whole content - even if we were only showing the first N characters, cutting in MySQL then stripping out HTML tags means we can end up with something way too short. Doesn't seem like the overhead is too much.

PR #251, opened 2015-06-17
[POC] New layer modules and themes

pupi1985:
Intention of the Proof Of Concept:
1. Get rid of qa_load_theme_class and the dynamic class construction (to call it some way)
2. Add more security by not allowing plugins to destroy the chain of calls by not calling a parent method
3. Remove from the theme all non-theme related behaviour
4. Manage layers as any other module (EG: getting rid of the $qa_layers global variable)

I know it pretty much destroys compatibility but the idea is not to merge this. It might also be buggy as the idea was to make something quick that would generate some discussion.

I'd create the PR for Q2A 2.0 but it doesn't exist yet :)


svivian:
There's a lot of good stuff in here! I like the idea of having the sort of "event" system where the base theme is checking if plugin layers need to be run. A few potential problems:

1. I'm not sure this offers the same flexibility as the current system. The `around_method` is kinda in a random place in each function. In short methods it's fairly useless while in longer methods there are multiple places users may want to insert content.

2. I'm far from sold on the sheer number of method calls required - an extra 3 in every theme method (and those call straight through to another so it's really 6).

3. Looping through the layer modules every time is overkill, but that can probably be solved by the theme pre-loading the list of methods to be overriden then doing `if (in_array($method, $preloaded))...`

We should also discuss the second point on your list. Should we prevent the chain of calls being destroyed? Sometimes a plugin may want to completely replace a function. But that does mean some plugins may not play well together.

So I'm not really sure of the best way forward...

pupi1985:
Thanks. Here are my replies:

1. Regarding flexibility, on one hand, the intention was not to make the core more flexible. In fact, this is related to your last question about not breaking the chain of calls. Being able to call (or not) a parent method gives you a lot of flexibility, IMO, too much. The question is: who controls the display of the different *components*: the theme or plugins? I guess it should be the theme as it is the skeleton in which the plugins should hang. Or put in a different way: the theme knows better how plugins would look better on it. I know there are plugins with complex UIs that are exceptions, though.

  On the other hand, there might be some flexibility improvements. The `around*` calls might enhance some situations in which a code should be placed between containers, e.g. `<div>{here}</div>`. Currently, when a method generates this HTML structure, it is impossible for a plugin to add code in that position. The plugin would need to override the whole method. Doing so is risky as the plugin might be overriding assuming the theme has a similar method to the base theme while it might not. Not to mention what would happen if 2 plugins override the same method in this way (security).

  I agree on the fact that in many methods the `around*` calls seem to be unnecessary. However, this is the base theme, i.e., other themes might find them useful. Anyway, I also see the same downside of these calls about having many places in the same method that could be *hooked*. The only way I can imagine this could be fixed is adding more hooks (methods).

1. You're correct. In fact, there are 1 array definition, 6 method calls and 2 string concatenations per method. All these multiplied by 166. The 6 methods calls could be turned into 3 by copy pasting some code, if necessary. Also, having to actually call a method in a plugin results in a couple of additional function calls... but considerably fewer in comparison. The thing is [this black magic](https://github.com/q2a/question2answer/pull/251/files#diff-598702d192702ef831f889eff9ca5076L1787) seems to be worse.

  The thing here is how many method calls does each of the following cost:
  *  A `file_get_contents`
  * `preg_replace` and `preg_match` in a short string
  * `preg_quote` x 6
  *  Array creation and iteration
  * `preg_match_all` in a usually long or very long string
  *  Many `substr_replace` (I give you the `array_reverse` for free :stuck_out_tongue:)
  * `eval`
  *  Others: `array_reverse`, many assignations and concatenations... and all these multiplied by each layer

  Probably, the number one performance killer is the `preg_match_all` as it can not even compile and cache the regex because it changes in each iteration. It is even worse as it seems to be case-insensitive. Not sure why...

  I did some tests here but were inconclusive. Both approaches on similar plugins took the same time (only taking into account theme PHP processing) but I only tested them using several plugins. Making a stress-test is not as easy as adding a `for` loop here.

1.  That is correct. In fact, I didn't focus on that because I knew some caching mechanism could be applied. That could be done in many ways: explicitly registering a method by means of a `register` function, automatically by inspecting the class and looking for some particular methods or maybe even with annotations.

As a side note, I had researched for some AOP libraries that would work with PHP 5.1 to implement these function calls in a nicer way but it was a failure. Most of the currently supported ones start at PHP 5.3+. I don't know what the right way is here. Maybe we could find (steal? :suspect:) something interesting from WordPress or Vanilla... but all I know about them is that they exist.

svivian:
The "around" methods don't feel like they will work in practice. If authors need to add code in the middle of functions then splitting the methods into smaller, fine-grained methods would be better (while keeping the before/after calls).

I took a look into Aspect Oriented Programming, can that actually be done with raw PHP? It sounds like the implementation is very convoluted.

---

I've been doing some research and found three other possible solutions:

1. Use the Decorator pattern. Essentially this involves each layer being passed into the constructor of the next one, and instead of `parent::method()` you call `$this->parenttheme->method()`. It would pretty much work identically to the current system but without the eval nastiness.

2. Use the Twig templating engine. For themes as a whole this may be simpler for a lot of people since you start with HTML code instead of HTML in PHP strings. Layer plugins could add their own overrides for certain blocks, and we can load & combine these.

3. Instead of directly overriding theme methods, allow authors to modify `$qa_content` before it's passed to the theme, and also allow the data in `$qa_content` to dictate what is output a little more. (For example, a 'tags' array key could provide additional attributes for each element, CSS/JS elements could be easily added to.) This covers a lot of the use cases of layers, and could be combined with your before/after approach.

pupi1985:
Regarding AOP: yes, the libraries I've seen around for PHP seem to provide too many features. Thinking in performance it would be better to avoid them. Maybe the before/after approach is enough in most cases and should work considerably faster.

Regarding the alternatives suggested:

1. I think if we're going to break the backwards compatibility from one day to the other with all plugins then better use the best approach. As I don't like the idea of plugins calling the parent method, I don't think this is the best approach. However, from a birds eye view, I think there could be some kind of transition implemented. Internally the core could use the decorators, however, before doing so it could replace parent calls and those variables, e.g. `QA_HTML_THEME_LAYER_URLTOROOT`, and then `eval`uate the plugins' code.
  I'd need to make a new POC to be fully convinced that this transition is possible. If it is then I think it is a good way to go.

1. I know nothing about Twig but I can use my imagination. From the page I see it requires PHP 5.2.4, which is not so far away from 5.1.6. I'm trying to imagine how switching to Twig would affect the base theme and allow plugins to work. Maybe there would be many template files (considerably fewer than the amount of methods in a current layer module)? There would be a set of files that would be part of the base theme. The theme would load those files but, while loading them, it would load and inject code from plugins which will read that code from template files too.
  If that is the case then it seems require a full rewrite of the base theme and split it into templates. Although I don't know anything about Twig, I don't think it would be possible for existing plugins to properly execute in this schema. However, it seems to be pretty flexible.

1. This is somehow what the `initialize()` method does. I had thought of a similar approach but I discarded it mostly because of a traversal issue, to put it some way. Having the methods that receive a `$post` as a parameter is essential, IMO. This saves a lot of coding. It has the downside that people end up not using it properly and adding SQL queries in methods call many times per page. Despite performance, the simplicity is key in order for plugins to flow. If they become to difficult to implement then nobody would take the time to do so.
  Regarding adding CSS/JS to the `$qa_content` array I think it won't be good. The array will be unreadable and huge. Probably I'm wrong with this as it might not be the intention to do this but I would like the content array to eventually have no CSS/JS/HTML at all. I understand the downsides of this and how the core generates hardcoded HTML and sends it to the theme but that can't last forever. I think the Holy Book somewhere says *And God separated the M from the V and He saw it was good*.

So I don't like approach 3, 2 would require me to learn Twig and also a rewrite of the base theme. Maybe it's worth investing some time in playing with approach 1 during July.

svivian:
Re: Twig, here's a simple example: https://gist.github.com/svivian/3dc47f0d701eb39fde1c
Typically you'd have a global master template for the "shell" design, then one template per view type (e.g. one for questions, one for question lists, one for user lists, etc). Then for our purposes we'd have a bunch of "blocks" (which can be nested) that could be overridden by plugins.

So the base theme wouldn't have many files, just lots of blocks in a few files. Custom themes should just be able to use one file containing all the blocks they wish to override. When I get time I'll put together a branch with a realistic example.

Re "CSS/JS elements could be easily added to" in 3, I think I meant adding/changing script URLs instead of passing HTML into the theme with script tags - for example allowing developers to change jQuery to a CDN version more easily or adding the JS file from their plugin.

In short, the idea is to make the `$qa_content` variable data only and be able to control the HTML output through it, but without having HTML in there.

svivian:
Also, in order to evaluate all these options, I guess we need to decide exactly what should and shouldn't be able to be modified by themes/layers. Do we only want people to add HTML and change data, or literally change any piece of HTML?

For example, take the `q_list_item` method. With your before/after/around system you cannot change the tag from a `<div>` to a `<span>` or `<li>`. Probably the same for my 3rd approach above (unless we added some convoluted system for specifying the tags within the `$qa_content` array).

But does that matter? Are plugins doing that currently, and are they likely to want to? It can mostly be mitigated with CSS but there are some edge cases such as the user/tag lists using tables which are quite inflexible.

pupi1985:
> Do we only want people to add HTML and change data, or literally change any piece of HTML?

I think themes should be able to change whatever they want. Although this could result in "non-standard" themes it would result in more creative designs. As long as themes are created in such a way that plugins can properly hook to them, that'd be perfectly fine.

On the plugin side, I think the opposite. They shouldn't be able to change the general structure of the theme. I think they should only add stuff, always making the least amount of assumptions about the theme.

Regarding the Twig example, it looks nice. However, I'm still not sure how the PHP file that would load the template would look like. I mean, if the idea is to add blocks in the template then there should be more or less a one-to-one relationship between Twig blocks and the current methods in the base theme. The new base theme should have almost no methods. So I wonder if there will be a processing loss, I mean, I guess theme/plugin developers won't be able to write PHP but rather Twig.

So this makes me think that the `qa_content` array would have to be pre-processed before processing the templates. So a lot of `for` loops will have to be done to place whatever code is needed by the theme. This is something the current base theme does for us and in each step it calls a convenient hook. As I see it we'd be losing that and I think that's a powerful feature.

Probably, I should wait to see what you have in mind with Twig before going on with this. But don't worry, I have some other radical ideas to work on in the meanwhile :)

svivian:
> On the plugin side, I think the opposite. They shouldn't be able to change the general structure of the theme. I think they should only add stuff, always making the least amount of assumptions about the theme.

OK that's a fair point. Doing that would make this whole system super easy since we would only ever have one "extension" of the base theme (the selected theme itself) and plugins could use something like your before/after solution. I'll look into what types of layer plugins there are now and how all of these solutions would affect them.

> I'm still not sure how the PHP file that would load the template would look like.

If it helps, here's what I had in my test file:

    $loader = new Twig_Loader_Array([
        'master' => file_get_contents('tmpl/master.twig'),
        'plugin1' => '{% extends "master" %}' . file_get_contents('tmpl/plugin1.twig'),
        'plugin2' => '{% extends "plugin1" %}' . file_get_contents('tmpl/plugin2.twig'),
    ]);

(In reality, the array passed into the function would be generated from the plugin list and the extends added automatically.) Twig itself creates 'compiled' PHP files that are similar to the base theme we have now, except all the output is in one function.

> I mean, if the idea is to add blocks in the template then there should be more or less a one-to-one relationship between Twig blocks and the current methods in the base theme.
> I guess theme/plugin developers won't be able to write PHP but rather Twig.

Yes for both of these.

> So this makes me think that the qa_content array would have to be pre-processed before processing the templates. So a lot of for loops will have to be done to place whatever code is needed by the theme. This is something the current base theme does for us and in each step it calls a convenient hook. As I see it we'd be losing that and I think that's a powerful feature.

Can you be more specific? You can do for loops and other control structures in Twig. (You can set variables and so on too, though that's usually best avoided in themes - the whole point is to output HTML, not to process more data.)

pupi1985:
I had to take a closer look at Twig before going on with this. So far, I'm not so sure the approach you suggest will satisfy our needs.

IMO, the ideal solution would have these features:

 1. A layer should depend on a small (ideally one) amount of (template) files
 1. The output of a plugin will not interfere with other plugin's output, i.e., there shouldn't be any need to call a parent in a chained way
 1. It should be possible to add code before and after certain tags (`{here}<head>{here}</head>{here}')
 1. Themes would create the skeleton and plugins won't be able to remove code, only add
 1. Variable (context) naming needs to be local to the layer (variable names from a layer shouldn't interfere with variables from other layers). This one is must
 1. Parsing dependent files (could be template files) should be minimized
 1. If the idea is to keep the `qa_content` array, it needs to be traversed once and allow plugins to hook on the method they want to add code to. E.G.: in order to add data to answers in a list I don't like the idea of navigating to the answers from the root of the `qa_content` array and iterating over them. I'd rather hook on a "print_answer" method/block/whatever.
 

Twig used in the way you suggest would comply with 1, 3, 6 but wouldn't comply with 2, 4, 5, and 7. I think it is not good enough. However, I've been thinking on a related Twig-dependant approach that would comply with 1, 2, 3, 4, 5, 6 and 7. I guess the only performant approach to comply with 7 is keeping the traversal of the `qa_content` array in plain PHP, as it nowadays work.

In the scenario I've analyzed that complies with 1-7 you need to register layers with their paths and contexts (variables they set). Templates don't need any kind of processing (as the "extends"). Then set a custom Twig extension that does some magic and send the context of each layer (ideally you'd send just the layer references which will have a `context` member). Then just render the main template (the theme).

The custom Twig extension creates a `q2a_layers('block_name', params)` function that is accessible from templates. So the theme template would call `{{ q2a_layers('head') }}` and then a function in the custom Twig extension will call each layer's template and execute that specific block assigning the context of that layer only. So this way plugins don't interfere with each other.

This approach also provides support for callbacks on the layer. Suppose you'd like to execute some PHP code as you can do currently by implementing `q_item_title($q_item)`. It could be possible to have the Twig extension call the layer `q_item_title($q_item)` method (even sending the parameter). Probably not performant but flexible. You can check it out here: https://github.com/pupi1985/question2answer/commits/patch-60-2

Let me know your thoughts on this.

svivian:
Can you give me a quick overview of your patch-60, and the files to look at? The diff is too large to load.

pupi1985:
I think you meant patch-60-2 (in the link above). Sorry, I was dumb and lazy. Dumb for originally branching from 1.7 and lazy for not correcting that. Check that branch again. The commits are simple and open ended... but will give you an idea of what I'm talking about.

PR #254, opened 2015-06-18
Update format.php

q2apro:
added semicolon for javascript, was forgotten, see here: http://www.question2answer.org/qa/20452/necessary-core-changes-for-correct-javascript-syntax-5-1-5-4?show=28220#c28220

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -1470,7 +1470,7 @@
 			$funcscript[]="\tif (e) { var d=(".$sources."); if (first || (e.nodeName=='SPAN')) { e.style.display=d ? '' : 'none'; } else { if (d) { $(e).fadeIn(); } else { $(e).fadeOut(); } } }";
 		}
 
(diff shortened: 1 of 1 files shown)


svivian:
I remember that question from a while back (I even posted in it!) but I don't see why that line needs a semicolon.

What is the error you get when minifying the JS? To be honest I'm not following why you are doing that, if you really want to minify the HTML/JS you should add a space after each line in the output function.

q2apro:
I am just fastidious here :)

svivian:
But what is the *problem* that you are encountering?

In a comment on Gideon's answer you said you're minifying the JS with an external tool but that doesn't make sense because that JS is generated dynamically via Q2A. And when I remove the spacing from the `output_raw` function (to 'minify' it dynamically) I don't get any issues.

q2apro:
Alright, then we leave it. I actually cannot remember if this missing char was a problem as well.

PR #255, opened 2015-06-20
Update question-view.php

q2apro:
Do not show answer button if user cannot answer anymore by spam limit


svivian:
Thanks for the PR. However, I think it makes sense to keep the answer button shown, otherwise it's not clear to the user why it disappeared (they do not see the "too many answers received" message).

PR #257, opened 2015-06-21
Update qa-widget-activity-count.php

q2apro:
added x percent of questions answered

### DIFF
--- a/qa-include/plugins/qa-widget-activity-count.php
+++ b/qa-include/plugins/qa-widget-activity-count.php
@@ -39,6 +39,14 @@ class qa_activity_count
 		$this->output_count($themeobject, qa_opt('cache_qcount'), 'main/1_question', 'main/x_questions');
 		$this->output_count($themeobject, qa_opt('cache_acount'), 'main/1_answer', 'main/x_answers');
 
+		// added x percent of questions answered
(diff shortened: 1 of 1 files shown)


svivian:
Interesting idea, but I don't think all sites will want to 'broadcast' their answered rate (in case it is low). So I don't think this is a good fit for the core. (Also it breaks on the SnowFlat theme.)

Since it's a very simple plugin other site owners can easily create their own with little effort. Thanks for the PR though.

gurjyot:
Actually this was nice one, it's good to know the percentage of questions answered on a website so admins focus on work. 
Secondly, the widget in SnowFlat can be changed to accommodate this, currently there are 4 boxes, instead make 5 lines - each representing one thing.

svivian:
Yes it's useful for admins to know the percentage, but for users not so much. Perhaps this could be incorporated into /admin/stats.

gurjyot:
Yes you also have a point Scott. But I wonder, if percentage is not to be revealed to users then why it is done in Stackexchange Forums. And most importantly, only new forums have this, the old and shining ones (like Stack Overflow) doesn't show that widget.

http://softwarerecs.stackexchange.com/
http://robotics.stackexchange.com/

svivian:
Those sites are in "beta" so the idea is to show that to users so they can go and answer more questions. Like I said, I think some sites would want this, but there are no options for built-in modules. Perhaps this can be converted to a regular plugin.

PR #259, opened 2015-06-22
Added admin option

q2apro:
Admin option to dis/allow anonymous to specify their name

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -1646,11 +1646,13 @@
 	$inname is from previous submission/validation. Pass $fieldprefix to add a prefix to the form field name used.
 */
 	{
-		$fields['name']=array(
-			'label' => qa_lang_html('question/anon_name_label'),
-			'tags' => 'name="'.$fieldprefix.'name"',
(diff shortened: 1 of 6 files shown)


svivian:
Sorry for the delay, just getting to this. Is there a particular reason we might want to disallow entering names for anonymous users?

For the closing in single field, as I mentioned on the Q2A site this would probably be better done via JavaScript. However this will be a good fallback if JS is not available for whatever reason. Are you able to update your commit to remove that colon? There should be a way to do that rather than add an extra commit (rebasing, maybe?)

svivian:
Pulled the Close Notice/Duplicate commit into 1.8. Thanks.

I left out the other commit as I don't see a reason for disallowing names.

q2apro:
Check the q2a forum. It has been asked so many times... even if you "don't see a reason" ;-)

svivian:
I don't see any questions on Q2A about it except yours, do you have some other examples? There is also [this question](http://www.question2answer.org/qa/25062/fault-with-request-a-name-for-anonymous-posts) where Gideon talks about adding it.

PR #273, opened 2015-07-21
[Bugfix] Fixed infinite loop and cleaned the core from some JS

pupi1985:
This works:

```php
qa_set_display_rules($qa_content, array(
	'target1' => 'source1',
	'target2' => 'source1',
));
```

While this doesn't as it seems to generate an infinite loop:

```php
qa_set_display_rules($qa_content, array(
	'target1' => 'source1',
));

qa_set_display_rules($qa_content, array(
	'target2' => 'source1',
));
```

The first commit fixes the issue and removes a lot of JS from the core, which can become very long sometimes!

The second commit suggest a different way of handling the source/target approach:

```php
qa_set_display_rules($qa_content, array(
	array(
		'source' => 'source1',
		'target' => 'target1',
	),
	array(
		'source' => 'source1',
		'target' => 'target2',
	),
));
```

The only pending thing would be to deprecate the old format.

### DIFF
--- a/qa-content/qa-page.js
+++ b/qa-content/qa-page.js
@@ -175,3 +175,29 @@ function qa_ajax_error()
 {
 	alert('Unexpected response from server - please try again or switch off Javascript.');
 }
+
+function qa_display_run(source, target, first) {
+	var sourceEnabled = source && (source.checked || (source.options && source.options[source.selectedIndex].value));
+	if (first || target.nodeName === 'SPAN')
+		target.style.display = sourceEnabled ? '' : 'none';
+	else {
+		if (sourceEnabled)
+			$(target).fadeIn();
+		else
+			$(target).fadeOut();
+	}
+}
+
+function qa_display_rules(first, rules) {
+	for (var i = 0; i < rules.length; i++) {
+		var rule = rules[i];
+		var source = document.getElementById(rule.source);
+		var target = document.getElementById(rule.target);
+		if (source && target) {
+			qa_display_run(source, target, first);
+			$(source).click({source: source, target: target}, function (e) {
+				qa_display_run(e.data.source, e.data.target, false);
+			});
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


svivian:
In what situation does the infinite loop occur? Is this in Q2A or from userland calls (i.e. a plugin)?

pupi1985:
I don't remember well but I'm sure the infinite loop was generated on the browser. It was a matter of how `onclick` callbacks were queued. According to the example I added, it seems to happen when having the same source and 2 different targets getting called in two different `qa_set_display_rules` functions.

svivian:
Sorry if I wasn't clear, what I was asking is when does that situation of having "two different `qa_set_display_rules` functions" occur in Q2A. It's not really important, I was just interested. The fact is, it *could* occur. I'll check this out on Monday and merge.

pupi1985:
Oh OK. I haven't seen it happen in the core itself. But it *might* happen when using plugins. Maybe 2 plugins using that function in the same page. I actually haven't faced that until I decided to take a look to the weird-looking `qa_set_display_rules` function :)

svivian:
Checked this out and it didn't work for me. Looks like it is not taking account of the fact that the source can be inverted (showing elements when the checkbox is *not* clicked). See for example `!dodelete` in `pages/admin/admin-pages.php`. You probably want to keep the first loop that builds `$keysourceids`.

svivian:
Played around with this a little and ended up fixing the infinite loop with minimal code so thought I may as well commit that. Are you able to redo your changes for cleaning up the JS? I think you can use mostly the same idea, you probably just need to store the raw source as passed into the function (e.g. `!dodelete`) as well as the actual element(s) we need to check from `$keysourceids`.

svivian:
Found a solution for this that cuts down the JS but preserves the flexibility. Thanks for the headstart with this!

PR #279, opened 2015-08-10
SnowFlat minor css modification for Login avatar box on navbar

q2amarket:
Fixed minor spacing issue at the topbar login box. (Avatar and user-name)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 # Ignore config file in development
 qa-config.php
+.idea
--- a/.idea/scopes/scope_settings.xml
+++ b/.idea/scopes/scope_settings.xml
@@ -1,0 +1,5 @@
+<component name="DependencyValidationManager">
+  <state>
+    <option name="SKIP_IMPORT_STATEMENTS" value="false" />
+  </state>
+</component>
--- a/.idea/scopes/scope_settings.xml
+++ b/.idea/scopes/scope_settings.xml
@@ -1,5 +1,0 @@
-<component name="DependencyValidationManager">
-  <state>
-    <option name="SKIP_IMPORT_STATEMENTS" value="false" />
-  </state>
-</component>
--- a/qa-theme/SnowFlat/js/snow-core.js
+++ b/qa-theme/SnowFlat/js/snow-core.js
@@ -84,4 +84,14 @@ $(document).ready(function () {
 		}
 	});
 
+
+    /**
+     * Toggle nav-sub (for admin and above)
+     */
+    $('.qam-nav-sub-toggle').click(function(){
+        $('.qa-nav-sub').slideToggle('fast');
+        $(this).toggleClass('active');
+    });
+
+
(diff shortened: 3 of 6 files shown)


svivian:
Can you update your PR to remove the 2 unrelated files? Thanks.

q2amarket:
Sure..., just getting bit trouble in another PR

svivian:
There's the `.gitignore` and there was something else in a `.idea` folder but I think you've removed that.

But the diffs for the `qa-styles.css` and `qa-theme.php` are messed up too. The CSS looks to have the indentation changed to spaces instead of tabs, and (at a guess) the PHP file has had the line endings changed? (Not sure how that happened though since the repo is set to normalise them...)

q2amarket:
@svivian Oh yeah, I have removed those files in fact.

Usually, I try to follow the [Coding Standards] and so I have set up 4 space for indention and probably my IDE has changed it. I also setup some spacing for phpdocumentor so if you see phpdoc blocks for all methods they have certain spaces between tag and parameter/description. This is all because to make doc-block more readable.

-------
**Help Required**
I am not so much handy with pull request so just wonder either you got my another request for Nav-Sub toggle modification for Admin and above users. 

**Commit:** https://github.com/q2amarket/question2answer/commit/a649ea65329cd51f4cdc6ebe433df2325d2b4cd8

**Suggestion:**
Since this is not on `gihub`, I would suggest to keep indention with 4 space for all code that is how it will looks the same indention with any text editor to everyone. But if you want I can try to change it to the tab indention, I believe, you too should also able to do in your IDE.

[Coding Standards]: http://framework.zend.com/manual/1.12/en/coding-standard.html

svivian:
Any good IDE/editor can automatically detect the spacing used in a file and adjust settings accordingly. Or maybe you can set it on a per-project basis. I'm not going to get into a tabs v spaces debate but the whole point of tabs is so you can make them look how you like. Anyway, that's the coding style used on the project so it should be followed.

BTW the coding standards you mention are for Zend Framework and are not anything official. Our coding guidelines are outlined in the [contributing file](https://github.com/q2a/question2answer/blob/master/CONTRIBUTING.md).

q2amarket:
I am using PhpStorm and Netbeans and both has been set with 4 spaces for a tab. So probably that is changing all tabbed lines with spaces. However, I have no issue to keep tab for indention for Q2A and I will do it from next update.

I am aware of the coding standard is specifically designed by Zend and it is not any official stuffs. Since we are working on open source, declaring or defining our own standard or following anyone can keep code cleaner and readable to the devs. If you don't prefer to follow zend than you can define own for Q2A.

BTW, from next updates I will try to keep the file as it is without re-formatting it

svivian:
There was a lot of unrelated stuff in this PR so I've taken the bits I think were relevant for the reported issue into 10912fb4dd328bcde308dc6cc9f488c0888f63b7

PR #292, opened 2015-09-25
Fix for user profile page

amiyasahu:
The profile data overlaps the current div , this commit fixes this 

Current view - 
bug

After fix - 
solved


svivian:
Thanks! This has been merged into the dev branch.

For future reference: pull requests should be made to the dev branch (unless it's a big change like a new feature, then it should be the 1.8 branch). Also for theme changes, if they are applicable to the other themes you can update those too if it's not too much trouble.

amiyasahu:
Thank you @svivian for accepting the PR . Sure I will check for other themes .

PR #296, opened 2015-10-02
Fixes #294, fixes #295: use local webfonts only.

### LINKED ISSUE #294: SnowFlat does not load web fonts when running on a https URL (TLS/SSL)
I noticed that when I moved my Q2A instance from http to https, the title font changed. Instead of Ubuntu, it used some sans-serif replacement that looks roughly like Arial. The reason was easy to spot in [`qa-theme/SnowFlat/qa-theme.php`](/q2a/question2answer/blob/4fa0914748eeac240206eb45356f35242f564cf1/qa-theme/SnowFlat/qa-theme.php#L70):
````css
$this->content['css_src'][] = 'http://fonts.googleapis.com/css?family=Ubuntu:400,700,400italic,700italic';
````
There is a trivial fix: Use a protocol-agnostic URL by dropping the `http:` part:
````css
$this->content['css_src'][] = '//fonts.googleapis.com/css?family=Ubuntu:400,700,400italic,700italic';
````
But instead of doing this, I decided that I do not want to force my users to connect to a Google service, and instead installed the font files locally. I am preparing a merge request for this.

### LINKED ISSUE #295: SnowFlat uses Google web fonts even though the fonts are redistributable
This is a privacy issue. Q2A admins might well prefer to run a tracker-free website. Why make connections to a Google service if you could just as well include the Ubuntu fonts in the theme? Their license permits this. I am preparing a pull request for this.

The way web fonts are loaded has to be changed to independent reasons anyway (see #294).

pietsch:
I tested this change with all the major browsers. The look is exactly like before. The is no need to provide formats other than TTF <http://caniuse.com/#search=ttf>.

Do keep the license file in there. The license also requires that you do not rename the font files.


svivian:
Hmm, took a look and I think if we're choosing one format, WOFF is better to use since it has a smaller file size (it's half the size of TTF). Support is around the same: http://caniuse.com/#search=woff (in fact if you ignore "partial support" then WOFF is higher).

svivian:
Decided to go with WOFF as it's more efficient. Also added a switch in the theme so people can still use Google Fonts if they like. See 097b73b

Thanks for the headstart with your PR!

PR #297, opened 2015-10-12
Q2A user levels with Wordpress integration

arioch1984:
Hi!
I use your great Q2A product and I really appreciate your work! 
So thank's a lot for it!

Using it with WP users external DB (great integration!) I found a problem: when users log in Q2A thought their WP account Q2A relative users level won't be stored on Q2A DB, instead they are loaded dynamically during login.

This is a problem for me because I had to implement some theme / plugin functionalities based on Q2A user level (moderator) and I had to catch this trought a DB query for performance reasons.
So I decided to modify externalisers-wp.php in my Q2A version to write on Q2A DB users levels logged thought WP login.

I hope this help and you like to accept my pull request.

Best regards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 # Ignore config file in development
 qa-config.php
+/.idea
--- a/qa-include/util/external-users-wp.php
+++ b/qa-include/util/external-users-wp.php
@@ -41,33 +41,49 @@
 		);
 	}
 
+	function qa_get_logged_in_user() {
+		$wordpressuser = wp_get_current_user();
 
-	function qa_get_logged_in_user()
-	{
-		$wordpressuser=wp_get_current_user();
-
-		if ($wordpressuser->ID==0)
+		if ( $wordpressuser->ID == 0 ) {
 			return null;
+		} else {
+			if ( current_user_can( 'administrator' ) ) {
+				$level = QA_USER_LEVEL_ADMIN;
+			} elseif ( current_user_can( 'editor' ) ) {
+				$level = QA_USER_LEVEL_EDITOR;
+			} elseif ( current_user_can( 'contributor' ) ) {
+				$level = QA_USER_LEVEL_EXPERT;
+			} elseif ( in_array( 'iam_moderator', (array) $wordpressuser->roles ) ) {
+				$level = QA_USER_LEVEL_MODERATOR;
+			} else {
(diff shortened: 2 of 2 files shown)


svivian:
Hi @arioch1984 finally looked into this. Unfortunately I don't think I can accept this PR.

As far as I can see, Wordpress does not have a 'moderator' permission, is that a custom permission on your setup? That's not something which belongs in core Q2A.

I can't see how the latter part of the code is helping, because it's still doing a query to get the user levels. And there isn't a mechanic for updating the user level. If they are changed from Admin to Editor in Wordpress, they will still appear as Admin in the Q2A database.

Finally, the coding style is incorrect, for example we don't use spaces inside parentheses (see the [contributing](https://github.com/q2a/question2answer/blob/dev/CONTRIBUTING.md#coding-style) file for details).

arioch1984:
Hi @svivian! 
Thank's for your reply.
You are right and I'm really sorry for that.
The moderator role you read on my code is intended as a custom role because the code release in my pull request is a wrong (perosnal) branch.
My real goal is to:
- save Q2A user level imported from WP on DB and not only load level on the fly while system read user role from WP (this is because in this way you could implement other functionalities in theme and plugins testing user Q2A level from Q2A DB directly)
- add the feature to map WP custom user roles to some Q2A user level (es. ham_moderator to QA_USER_LEVEL_MODERATOR)

Of course in Q2A this is possible by using qa-external-users.php but in this way you have to rewrite entire default WP integration and then write your custom code on it.

Last but not last, sorry for my code format, I didn't see the contributing file details, now I would like to read it.

So may we close this pull request and then, when I have re formatted my code, may I try another pull request from the right branch this way?

Thank's in advantage and sorry for my horrible english.
Best regards.

svivian:
When you say "save Q2A user level imported from WP", what are we saving? All we are doing is loading is loading from the Wordpress tables instead of the Q2A tables, we don't save anything there.

Themes and plugins can already call the user functions (such as the [ones listed here](http://www.question2answer.org/functions.php)) and Q2A will handle returning the WP details if you're using WP integration, or Q2A details if you're using the built-in user accounts.

For adding custom WP roles, that's not really doable in core Q2A since everyone has different customisation. Your best bet is to use the [single sign-on](http://www.question2answer.org/single-sign-on.php) but copy the code from `util/external-users-wp.php` to `qa-external/qa-external-users.php` and modify it to your needs.

PR #301, opened 2015-10-21
[BugFix] - Updated Google reCAPTCHA plugin and fixed dependency for allow_url_fopen configuration

### LINKED ISSUE #300: qa-plugin/recaptcha-captcha needs allow_url_fopen to be enabled
The qa-plugin/recaptcha-captcha needs the allow_url_fopen option to be enabled in the server's config, which is usually disabled by the hosting provider of the website.

Infos on this issue can be found here:
You can follow the instructions here for more information on this issue - google/recaptcha#36

Please note also, that the version of recaptcha shipped with q2a is outdated, and/or not the full version.

amiyasahu:
fixed #300


svivian:
Hi, thanks for taking the time to do this. Unfortunately, the latest version of Google's reCAPTCHA library requires PHP 5.3 but Q2A is still supporting 5.2. The library is only a wrapper around the API so it's not necessary to use the latest version.

If you're able to take the socket stuff and integrate it here that would be great. Otherwise I'll take a look at doing it in the near future.

amiyasahu:
Ok , Sure . I will take a look and try to integrate the Socket method validation here . Thanks

amiyasahu:
Closing this . I will send another PR with the fix

PR #303, opened 2015-10-23
[Improvement] - redirect the user if the slug is incorrect 

### LINKED ISSUE #253: Redirect URL if slug is incorrect
While debugging an encoding issue, I noticed the software does not notice when invalid url's are entered:

This page:
http://mathsgenius.co.za/qa/961/solve-%24x-2-2x-1-0%24

Can also be reached on:
http://mathsgenius.co.za/qa/961/solve-?x-2-2x-1-0?

And: 
http://mathsgenius.co.za/qa/961/solve-&%24x-2-2x-1-0%24

Probably because the 3rd segment of the URL is ignored. Might want to do some checking.

amiyasahu:
This fix checks for the question url . If it does not match with the
current URL then redirect to the correct one

fixed #253

### DIFF
--- a/qa-include/pages/question.php
+++ b/qa-include/pages/question.php
@@ -56,6 +56,11 @@
 		$question=null;
 
 	if (isset($question)) {
+		$q_request = qa_q_request($questionid, $question['title']);
+
+		if ($q_request != qa_request())
+			qa_redirect($q_request); //redirect to appropriate URL if the current request URL is incorrect
+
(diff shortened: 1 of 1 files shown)


svivian:
Thanks, merged into 1.8.

@yerbol89 the 400 Bad Request error is being generated at the server level, i.e. before Q2A is even run. So there's nothing we can do about it.

amiyasahu:
Thank you

PR #305, opened 2015-10-23
[Improvement] - Content focus when a answer or a comment is opened (Fixes #234)

### LINKED ISSUE #234: Won't it better to improve content focus when it is opened (Like a question or a answer)?
I was going through different QA cms and came across one which has some animation sort of stuff. When ever we click a link on that website then whatever opens, show some colour in the background to highlight the exact stuff and then it goes normal.
Here is a screencast for that, I click on a question and when it opens then it shows a light blue colour in the background and then vanishes to tell what exactly that link was. 

q2a3

This thing will be cool in Q2A, specially when a user click on a link (i.e answered or commented) on the homepage. This thing can also be used for wall posts and even messages (specially when notification system will be built).

amiyasahu:
This PR fixes #234


svivian:
Thanks for this and the other PRs. I made a small change to this one adding the `-webkit-` prefix. Also it didn't work on the Candy theme because Candy uses a background image for answers, so I've removed it from that theme.

amiyasahu:
Thank you @svivian .

PR #306, opened 2015-10-24
[BugFix] - Fixed the issue with user registration & confirmation when the user must confirm email is enabled 

### LINKED ISSUE #205: Issue with user registration & confirmation
When users are registered and it's required that they confirm their email, they won't be able to change their mail before email confirmation.

Scenario is that user registers with fake email account and when he is requested to confirm it, he decides to enter his real email. but he won't be able to change it.

amiyasahu:
Fixed #205


svivian:
Thanks, good work! Merged into 1.8.

PR #314, opened 2015-11-03
Block layout for users and tags page

### LINKED ISSUE #310: Users page design in mobiles is very cluttered.
Users page design on mobiles is very bad in SnowFlat. There is not sense of symmetry in that page. See to the pic below.

users

It would be much better to have users page something like this.

users1
users2

This design is much better then what we have in Q2A. And if you remove that sidebar in users page (as no one visit users page for having a look at sidebar) then this page will look even better.

amiyasahu:
Fixes #310

### DIFF
--- a/qa-theme/SnowFlat/qa-styles.css
+++ b/qa-theme/SnowFlat/qa-styles.css
@@ -31,7 +31,8 @@
 .qa-a-list-item:after, .qa-a-item-buttons:after,
 .qa-c-item-buttons:after, .qa-c-item-clear:after,
 .qam-footer-row:after, .qam-qa-list-meta-box:after,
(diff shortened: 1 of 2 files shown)


svivian:
I tested this and I'm getting an issue where the avatar increases the height and causes the floats to move to the wrong place:

screen shot 2015-11-14 at 23 02 06

Should be fixable using `display: inline-block`. But I'm wondering if there is something better we can do - maybe closer to the other themes where you have the avatar on the left then 2 rows on the right with username and points.

amiyasahu:
Ok . I will take a look into this

gurjyot:
Can you guys do something which can take benefit of complete page like in Donut theme or on SO?
Means hiding sidebar on Users page.

PR #317, opened 2015-11-10
Responsive adsence

### LINKED ISSUE #313: Improve basic-adsense plugin for responsive ads
Q2A 1.7 was all about responsive design and making everything working under mobile. But one thing was left completely untouched and that is ads. 
Right now, if I apply an ad to show just after menu then it shows perfectly, but when I open the website in mobile then that ad is not made to fit in mobile screen. Instead a half cut ad shows which is worst as user doesn't understand that what's the half ad is for and it even looks bad.

So please improve that plugin and provide some solid solution for placing ads. It's one of the most important plugins for users so make it worth to be in Q2A core, otherwise ditch this plugin and include a third party plugin which can take care of most ads type.

amiyasahu:
fixes #313


svivian:
This all looks fine except you spelt "Adsense" wrong in the CSS :)
Could you fix that and the commit message please?

amiyasahu:
Oops , My bad . I will do that .

amiyasahu:
I will make another PR with correct spelling

Issue #309, opened 2015-10-28
Clicking "Ask Question" button more than once creates duplicate questions.


gurjyot:
The problem is mainly with slow internet speed, if a user is submitting a question with slow internet speed, then he might click "Ask Question" button twice or thrice. The multiple clicking of that button creates duplicated questions of the original questions.

If there could be applied a restriction that once the button is clicked then it can't be clicked again, that can solve this issue. What do you say about this?

svivian:
When this happens they should get an error saying "your submission appears to be a duplicate". Not sure why that's not the case here.

gurjyot:
That's the problem, instead of showing an error it creates a lot of questions. Even in good internet speed if you try to click multiple times on that button then also it'll generate duplicate questions.

svivian:
OK looks like that duplicate detection is only on answers and comments currently. Definitely makes sense to add this to questions too.

amiyasahu:
But checking a currently posted question against all available questions would be a costlier process . Because if you are comparing the submitted content with all other answers or comments there will not be too many . But comparing with all the questions there may be 1000's . 

It can be possible check the duplicate questions which is posted in last hour only , or something like that . 

Or just disabling the ask button after it is clicked would be also a alternate solution (Only at UI level though) .

svivian:
Yes the case for this is when the user clicks the 'ask question' button several times in a short space. I think we'd just need to check the latest few questions (5 at most).

Disabling the button with JS on click would also be a good idea to add on top of that, since it will prevent extra hits to the server and extra checking. We can add that to all post types - questions, answers, comments, wall posts etc.

gurjyot:
Yes disabling button for multiple hits (for all post-types) would be nice idea and if questions check is also introduced then that will be like honey on icecream... ;)

phy25:
I don't think it is a good idea to directly disable submit button. If the network failed and for some reasons the button is not enabled again, the user won't be able to submit the content again unless backuping the content themselves and refresh the page. That's annoying.

Mr-Kumar-Abhishek:
Uhh.. Why not send an ID to the client for every question they want to ask . And verify that when the post is received by the server . If that ID is already posted reject it ? I am just saying haven't read the whole code of this repo
Sent via Micromax
[...]

PR #318, opened 2015-11-10
[BugFix] - restrict the user to post duplicate questions

amiyasahu:
do not allow user to post the duplicate question once more time . It
checks the last 5 questions and compares with the current one . If any
of them appears to be same the user will be warned and the question will
not be posted
Fixes - #309

### DIFF
--- a/qa-include/db/selects.php
+++ b/qa-include/db/selects.php
@@ -1692,7 +1692,29 @@
 		return $selectspec;
 	}
 
+	/**
+	 *  Return the selectspec to retrieve recent questions by the user identified by $identifier, where $identifier is a
+	 *	handle if we're using internal user management, or a userid if we're using external users. Also include the
+	 *  corresponding vote on those questions made by $voteuserid (if not null). Return $count (if null, a default is used)
+	 *  questions.
+	 *
+	 * @param null $count
+	 * @param int  $start
+	 *
+	 * @return array
+	 */
+	function qa_db_user_recent_qs_full_selectspec($count=null, $start=0)
+	{
+		$count=isset($count) ? min($count, QA_DB_RETRIEVE_QS_AS) : QA_DB_RETRIEVE_QS_AS;
 
+		$selectspec=qa_db_posts_basic_selectspec(null, true, false);
+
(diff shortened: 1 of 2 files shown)


svivian:
Is there a reason for the new function? Shouldn't `qa_db_qs_selectspec` cover what we need? You can call it with mostly blanks and it should be as simple as your new function.

    qa_db_qs_selectspec(null, 'created', 0, null, null, false, false, 5);

amiyasahu:
An small correction to your suggestion - full post is needed as we need the content of the post 

```
qa_db_qs_selectspec(null, 'created', 0, null, null, false, true, 5);
```
The result will remain same by the way you have provided . Initially I also thought this way . 
But with this , the code does an extra join with the user table which is not necessary . For that shake It costs me a extra function .

The existing code of `qa_db_qs_selectspec` will call the qa_db_posts_basic_selectspec like this - 
```
$selectspec=qa_db_posts_basic_selectspec($voteuserid, $full);  //$user variable becomes true by default
```
But the new function calls like this with false for the user variable 
```
$selectspec=qa_db_posts_basic_selectspec(null, true, false);  // false to the $user variable 
```

Which avoid the unnecessary join with the user table . If the tables were huge then it would make a significance difference ( what I believe ).

If we go for minimum code changes we can go with the existing function , but if we see the performance point of view it is worthy to add an extra function . 

What do you say ?

svivian:
Actually that join inside `qa_db_qs_selectspec` speeds up the query. This is one of the things Gideon told me when I took over Q2A development that I didn't know about MySQL. Basically, joining the table to itself like that means we do a fast select based on the table index. See the [last point here](https://www.percona.com/blog/2009/09/12/3-ways-mysql-uses-indexes/) for more info.

So you can call `qa_db_qs_selectspec` with `$full` set to true as you said, and it will be efficient.

amiyasahu:
Ok , Still I am not sure how can a extra join will improve the performance of the db query ( may be due to my limited knowledge ) .

I have created another PR #323 to avoid unnecessary commits with using the existing q2a functions . Both of them works in the same way 

I am closing this for now . Thank you :smile:

PR #323, opened 2015-11-15
[BugFix]-Restrict the user to post duplicate questions

amiyasahu:
do not allow user to post the duplicate question once more time . It
checks the last 5 questions and compares with the current one . If any
of them appears to be same the user will be warned and the question will
not be posted
Fixes - #309


svivian:
Thanks for this. FYI it needed an extra check for the question title too, because multiple questions can have identical content - when a user posts a title only and nothing in "More information for your question".

PR #324, opened 2015-11-15
Responsive adsense

amiyasahu:


svivian:
Hi amiyasahu, sorry but there are a couple of new problems with this PR:

1. The indenting is messed up. The Q2A code uses tabs for indenting, not spaces. You might need to set an option in your IDE/text editor. The first block (array) is also double-indented.
2. It looks like you have missed out the line from #317 that saves the ad unit ID. So I think that part is not going to work?

svivian:
By the way, you don't need to make a completely new PR when you make changes. You can make changes on your fork, do a rebase (for example squash the commits into one) and then push it back to your fork, and the PR here will update automatically. [This article](http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/) has a good explanation.

amiyasahu:
Oh my bad . Sorry I did not know about the force amending the PR . I will do the required changed in this PR and push it back .

gurjyot:
@amiyasahu and @svivian I would suggest not to do make this PR again. Basic Adsense plugin is too much basic to be used in 2015 when we need much more information rather than just displaying some ads.
A better way would be, to add Widget Anywhere plugin in Q2A core removing Basic Adsense. There are two most important reasons for this.
1. Since we would like to display different sized ads in different places and 
2. We also want to know that which particular ad is giving us most earning so that we do not remove it or change it's position.
Both of these things are not possible with Basic Adsense but both these can be done with Widget Anywhere plugin. 
So in turn improve Widget Anywhere plugin a little bit and use that in core instead of this old Adsense plugin.

svivian:
I've made the style changes and merged your code. Thanks for the PR.

amiyasahu:
Sorry for the delay @svivian , I was quite busy since last couple of weeks . Thanks for the efforts .

PR #342, opened 2016-01-03
prevent directory listings

DanielRuf:


svivian:
Thanks!

q2apro:
To be considered: http://stackoverflow.com/q/5932641/1066234

svivian:
@q2apro What are you referring to? The `Options -Indexes` has been added. If a developer is unable to use htaccess they should edit the config at the server level, or contact their host.

Listing the files isn't really a big problem. Q2A is open source so it's pretty obvious what files are there!

PR #349, opened 2016-01-08
Added a pseudo template engine

jblandry:
Added a pseudo template engine which permits to output html from external files.

Searches for files under templates/*.phtml under theme folder

### DIFF
--- a/qa-include/qa-theme-base.php
+++ b/qa-include/qa-theme-base.php
@@ -139,6 +139,20 @@ class qa_html_theme_base
 	}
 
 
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for the suggestion, but this seems like too specific a thing to be in core Q2A. It's more suited to custom themes/plugins. PHTML also depends on server config, it's not a standard part of PHP.

(By the way we are considering moving to a real template engine like Twig in the future, so that would solve this problem in a more complete way.)

Issue #340, opened 2016-01-02
password_hash with bcrypt instead of sha1(salt+password+salt)


DanielRuf:
Are there any plans to move from salted SHA-1 to password_hash and the other password_* functions available in PHP 5.5+ (password_compat from ircmaxell available as polyfill) with the PASSWORD_BCRYPT flag ( + salt + cost )?

svivian:
The problem is that we currently support PHP 5.2, and the polyfill only works from 5.3+.

However, I am planning to add a strategy for this in the next version (1.8). I'll add an extra field to the users table to hold the bcrypt password, then do a version check and if the site is running on 5.3+ migrate each user's password as they log in (and use bcrypt for new user registrations, of course).

At some point in the future we'll drop support for 5.2 and we can migrate the rest of the passwords.

DanielRuf:
Sounds good. This is how I would also do it (extra field and silently convert) like you describe it.

5.2 and some newer 5.* versions of PHP are officially not supported anymore. 5.3 should be at least be used as it supports important cryptography related functions.

If I find the time I can provide some pull request for this.

The field should be binary UTF-8 field with a length of 60.

svivian:
That would be great!

Re PHP versions, the problem is that "official support" unfortunately does not correspond with usage. There's still a significant percentage of sites using PHP 5.2 which we'd like to support - although it looks like that's [dropped off a fair bit](http://w3techs.com/technologies/history_details/pl-php/5) this year. Whenever Q2A 2.0 is released we'll certainly be requiring 5.3+.

DanielRuf:
For PHP 5.3+ include the functions from https://github.com/ircmaxell/password_compat

PR #351, opened 2016-01-14
[dev] use password_hash and password_verify

DanielRuf:
Please test it.

This pull request converts user hashes after a successful login, and uses bcrypt for all new passwords by default.

You need to add password_compat functions to question2answer to make it compatible with PHP 5.3+ (I work with PHP 5.6 and 7.0 on my local machine).

Worked here without any issues.

This pull request solves #340

Keep in mind that we do not need any salt as password_hash creates a safe hash by default.

### DIFF
--- a/qa-include/db/install.php
+++ b/qa-include/db/install.php
@@ -1429,13 +1429,13 @@
 					$keyrecalc['dorecalcpoints'] = true;
 					break;
 
-			//	Up to here: Verison 1.7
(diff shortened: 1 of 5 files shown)


DanielRuf:
@svivian Can you merge this and my previous commit https://github.com/q2a/question2answer/commit/363800c06c48a1b7ca7b236cd56dcd6b3f403e8d with the 1.8 branch?

PR #353, opened 2016-01-14
[1.8] use password_hash and password_verify

DanielRuf:
Please test it.

This pull request converts user hashes after a successful login, and uses bcrypt for all new passwords by default.

You need to add password_compat functions to question2answer to make it compatible with PHP 5.3+ (I work with PHP 5.6 and 7.0 on my local machine).

Worked here without any issues.

This pull request solves #340

Keep in mind that we do not need any salt as password_hash creates a safe hash by default.


DanielRuf:
@svivian Should I close this PR too or do you need it as a reference? Because there were some other changes in the 1.8 files which I had to backport to the current dev files in my branch.

svivian:
Since this is a major change it should be against the 1.8 branch. So this PR is fine and the other one should be closed.

However I'm not sure exactly what you mean about backporting to dev. There shouldn't be any need to move stuff between branches. In your fork you'd make the changes on the 1.8 branch and push the PR from there. I think you've done that correctly here.

DanielRuf:
Ok.

Initially I used implemented the password_ functions in a local copy of the current dev branch.

Then I copied these files to the 1.8 branch and reapplied the changes from the 1.8 branch.
This is what I meant. Sorry for the confusion. 

Makes sense, when there are many commits between dev and 1.8 that the files in the branches are also different. Just eanted to make sure that everything is ok with the PR.

svivian:
Hi Daniel, I've taken a look at this and it seems to be missing the backwards compatibility we discussed.

For support of PHP 5.3+ we'd need to include the [password_compat](https://github.com/ircmaxell/password_compat) library. Also, for 5.2 support we'd need to add a version check when updating accounts. To clarify, the process should work like this:

1. Update the db to add the new field (which you've done).
2. If we're on PHP >= 5.3.7, include the password_compat library where appropriate - you can use `!qa_php_version_below('5.3.7')` to check for that.
3. When creating accounts, if we're on PHP >= 5.3.7 use the new `password_hash` functions. Otherwise, do what we did before (`qa_db_calc_passcheck` etc).
4. When logging in, if we're on PHP >= 5.3.7 then switch the user's password to the new format. Otherwise continue to use the old method.

Hope that makes sense.

DanielRuf:
Ok, wil prepare some commits.

DanielRuf:
The commits https://github.com/DanielRuf/question2answer/commit/e0883c52482daf2d19c83188e774c470b21496df, https://github.com/DanielRuf/question2answer/commit/6f6f34573b852ec83b287fa6e702b123c7fdcd94, https://github.com/DanielRuf/question2answer/commit/f0108fe665d4c8e730f1ab44a36869e2449d6198 should solve 3 and 4.

DanielRuf:
Should we use `include_once lib/random.php` in the config file? And should we bundle random_compat with question2answer? If so, where?

svivian:
Currently we keep third-party code in the `qa-include/vendor/` folder, so something like `qa-include/vendor/password_compat.php` is fine.

Since even including that file requires 5.3 (due to the namespace), we'll have to do it inside the version checks i.e. at the places we actually use `password_hash`.

DanielRuf:
Ok.

Should I add the include statement to all files which use the password_ functions or some central file like the config file, so there are less checks on the serverside, if the file is already included?

svivian:
The best place would probably be in `qa-base.php`, you can put it after the JSON compatibility block.

DanielRuf:
Ok, pushed the commit to the branch right now. So all points https://github.com/q2a/question2answer/pull/353#issuecomment-172929235 are done?

svivian:
Yep, all looking great! I'll test and merge in later today.

svivian:
All merged in, thanks for the PR!

BTW I changed a couple of minor things. There was a bug in the SQL query (extra parameter key, see b29c76357326aaae0e430cefde5a46741c4c6fc9). Also the PHP docs suggest using a larger field for the passhash as it could potentially increase in future (2d36adfa30f6a0e906b58f98971e3dc7c624549f).

DanielRuf:
So this works now? Oops, seems I forgot to remove the extra SQL parameter.

PR #360, opened 2016-01-22
Fixed Security by adding tags

dodiaraculus17:
added a die after the require. Disabled Error Reporting.

### DIFF
--- a/index.php
+++ b/index.php
@@ -21,11 +21,11 @@
 */
 
 //	Set base path here so this works with symbolic links for multiple installations
-
+        error_reporting(0); //For Security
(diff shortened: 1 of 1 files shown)


svivian:
> added a die after the require.

Why? That's the end of the script, so unless I'm mistaken that does nothing.

> Disabled Error Reporting.

This actually won't work here as `error_reporting(E_ALL);` is set in `qa-base.php`.

But we shouldn't turn off error reporting. The problem of errors being displayed on the website is solved with the `display_errors` and `error_log` ini settings, not disabling them entirely. It's up to users/hosts to set those responsibly.

dodiaraculus17:
Thanks that makes more sense.

PR #369, opened 2016-01-28
Replaced registration URL for WordPress integration

IgorCode:
Q2A doesn't use `wp_registration_url` WP function for fetching registration URL so it bypasses `register_url` filter. This filter can be used for adding custom registration pages.

I added WordPress function `wp_registration_url` for fetching registration URL while still keeping compatibility with older WordPress (bellow 3.6) that do not have this function.


svivian:
Thanks!

Issue #370, opened 2016-01-31
PHP Question2Answer MySQL query error 0


q2apro:
I tried to migrate the uploaded files from the database table qa_blobs (1.6 GB) to the filesystem under admin/stats → Button "Blobs to Disk".

Result frontend: First it shows "0 of 29930 uploaded images and documents" then it appears "undefined"

error_log says:

> PHP Warning:  mysqli::query(): (00000/0):  in /qa-include/qa-db.php on line 197
> PHP Question2Answer MySQL query error 0:  - Query: SELECT blobid, content, format FROM qa_blobs WHERE blobid>=0 AND content IS NOT NULL

The undefined appears after ~10 seconds. It is not due to a CPU timeout, my hoster also wrote me that the RAM is sufficient. 

---

Line 197 of qa-db.php is:

	$result = $db->query($query);

sama55:
It would be a bug due to insufficient process. MySQL packet size (max_allowed_packet) error might have occurred on your DB server.

Try to below hacks.

app/recalc.php
case 'doblobstodisk_move':
...
	//$next=1+$blob['blobid'];
	$next=$blob['blobid'];

case 'doblobstodb_move':
...
	//$next=1+$blob['blobid'];
	$next=$blob['blobid'];

db/recalc.php
function qa_db_get_next_blob_in_db($startblobid) {
	//'SELECT blobid, content, format FROM ^blobs WHERE blobid>=# AND content IS NOT NULL',
	'SELECT blobid, content, format FROM ^blobs WHERE blobid># AND content IS NOT NULL LIMIT 1',

function qa_db_get_next_blob_on_disk($startblobid) {
	//'SELECT blobid, format FROM ^blobs WHERE blobid>=# AND content IS NULL',
	'SELECT blobid, format FROM ^blobs WHERE blobid># AND content IS NULL LIMIT 1',

q2apro:
Wow! Thanks for your help. The fix was to put the `LIMIT 1` at `qa_db_get_next_blob_in_db()`.

I have created a pull request: https://github.com/q2a/question2answer/pull/371

PR #371, opened 2016-02-01
Fix issue 370

q2apro:
https://github.com/q2a/question2answer/issues/370


svivian:
Thanks. BTW I also added `LIMIT 1` to the `qa_db_get_next_blob_on_disk` function too.

Should be part of 1.7.3, assuming Gideon hasn't started preparing the release already.

PR #377, opened 2016-02-06
fully support IPv6

DanielRuf:


svivian:
Also would you be able to squash the "support IPv6 using inet_pton" commits into fewer commits? At least, all the ones changing queries would be better as one commit rather than many. Mkes it easier for me to look through, thanks :)

DanielRuf:
Will address these the next days.

Unfortunately I made the commits here on GitHub. Would have to 

> I think we would need to keep the query parameter here and pass in the converted IP, i.e. createip=$ in the query and inet_pton($ip) on the next line. Otherwise it's going to mess up the query if there are special characters.

Special characters in the IP? I thought the other parameters are properly escaped.
Which commits exactly need this change?

> Sorry! I made a mistake with regard to NULLs. Looks like some of the fields allow NULL and some don't (I was only looking at the qa_posts table before). We should keep each field the same as it was, so just change INT UNSIGNED to VARBINARY(16) and leave the NULL/NOT NULL the same.

Theoretically there can not be NULL when we use the ROME_ADDR header which can not be forked. And if it would contain some invalid date, it would have to be NULL when inet_pton returns false in this case.

> We'll need to check if the eventlog table exists, because it's not on every Q2A installation

Ok, should be possible with some additional check, will have to write and test this.

> Similar thing here, if QA_FINAL_EXTERNAL_USERS is set then the users table won't exist. See the example in case 62 above.

Okay, will take a look at the code that you have committed.

svivian:
> Special characters in the IP? I thought the other parameters are properly escaped.

Where you have `... WHERE createip=".inet_pton($ip)." AND ...` you are putting the value direct into the SQL query, unescaped. So you could end up with the SQL as `... WHERE createip=abc' AND ...`

In Q2A parameters are escaped when you use the `$` or `#` placeholders and pass in the arguments to the `qa_db_query_sub` function. See the comment for `qa_db_apply_sub` in qa-db.php. (I do plan on replacing this with the built-in parameterized queries at some point. It's a bit of a pain with mysqli.)

> Which commits exactly need this change? Just the one?

Looks like most of them. Basically `INET_ATON($)` should be replaced with `$`, and the `$ip` parameter should be replaced with `inet_pton($ip)`.

> Theoretically there can not be NULL when we use the REMOTE_ADDR header which can not be forked. And if it would contain some invalid data, it would have to be NULL when inet_pton returns false in this case.

The reason some of them allow null is because that action has not happened. For example the `createip` cannot be null because someone creating a user account always has an IP. But the `writeip` can be null because when the user is created they have not written anything yet.

DanielRuf:
Ok, thanks for the feedback.

DanielRuf:
> The reason some of them allow null is because that action has not happened. For example the createip cannot be null because someone creating a user account always has an IP. But the writeip can be null because when the user is created they have not written anything yet.

Changed in https://github.com/DanielRuf/question2answer/commit/da046f78946c039a683ee5bbb9e4bef3e067e81b

svivian:
Regarding the eventlog, I've just realised that it was storing the IP as plain text rather than an integer. The plain string value is passed into the insert query, no `INET_` function. Since this is used by plugins we'll need to keep the behaviour the same but expand the size of the field - 45 characters is the max according to [this](http://stackoverflow.com/questions/166132/maximum-length-of-the-textual-representation-of-an-ipv6-address) so a `VARCHAR(45)` should do it.

Besides that everything looks great! I'll test and integrate it in the next few days.

DanielRuf:
> Regarding the eventlog, I've just realised that it was storing the IP as plain text rather than an integer. The plain string value is passed into the insert query, no INET_ function. Since this is used by plugins we'll need to keep the behaviour the same but expand the size of the field - 45 characters is the max according to this so a VARCHAR(45) should do it.

Fixed in https://github.com/DanielRuf/question2answer/commit/eb6e7e466abba9d787a2b0472b588aa1b91e8bf8

svivian:
This has now been merged.

There were a few errors right off the bat, so in future please make sure to test your changes locally so you catch the rudimentary mistakes.

The main problem was that it turns out the `inet_*` functions emit Warnings when incorrect data is passed to them (for some reason this is not mentioned in the manual!). This happens prior to the DB being upgraded (because it's passing in the old integer value instead of binary), and also wherever an IP is NULL. I haven't found a reliable way to avoid those besides just suppressing the warnings with `@` everywhere. You can check out the other fixes [on the branch](https://github.com/q2a/question2answer/commits/1.8).

Thanks for the work on this. Besides a few oversights it worked really well!

svivian:
Just remembered one other thing - in `qa_block_ip_match` the `ipv6_expand` function is being called with a second parameter of "true", but that function doesn't have a second parameter. Do we need to add a parameter to the function, or remove "true" from the function call?

DanielRuf:
> Do we need to add a parameter to the function, or remove "true" from the function call?

Remove it from the function call, this parameter is not used. I am not sure why this is there and what I did there in the past. It seems this was just in the codeparts which were related to IPv6, the code parts related to IPv4 did not have the extra parameter. Anyway, just remove the extra parameter in the function calls. Sorry for this.

> The main problem was that it turns out the inet_* functions emit Warnings when incorrect data is passed to them (for some reason this is not mentioned in the manual!). This happens prior to the DB being upgraded (because it's passing in the old integer value instead of binary), and also wherever an IP is NULL. I haven't found a reliable way to avoid those besides just suppressing the warnings with @ everywhere.

Right. Unfortunately my testdata here had no NULLs, totally oversaw it when working on this PR.

If it is this, we may check `if(isset()` in the code before running `qa_block_ip_match` and return nothing if it is not set to prevent this. Wouldn't this be a better solution than suppressing errors (which is never a good idea)?
http://php.net/manual/en/function.inet-pton.php#93501
http://ideone.com/Swu5tc

You already do this in some places with `if (isset($createip))` where error suppression is not needed if it is not set (NULL). Or we add an additional check something like `qa_is_valid_ip` which just test if `if (filter_var($ip, FILTER_VALIDATE_IP))` returns true, see http://ideone.com/Swu5tc

But I see you use error suppression quite often in your code, this might be sufficient how it is solved currently (using the error suppression symbol). Error suppression in PHP is generally not a great thing anyway ;-)

Errors should be generally properly handled or throw an (own) exception which you can catch easily. See http://stackoverflow.com/a/960288/753676

This would be definitely a good idea in the future (own exceptions and handling errors properly instead of suppressing them) to avoid issues with changes in PHP (it nis not so rare that errors and notices will become fatal errors in later PHP versions and produce more problems) and allow debugging (suppressing them will make it impossible to debug them). :-) Just my two cents about using the error suppression symbol. `isset` is also faster in most cases than throwing an error but suppresing it.

DanielRuf:
Huh, but why varchar instead of varbinary? inet_pton creates a binary string and there is also the binary conversion. Varchar may not contain binary data or produce issues. Varchar is jsut for saving REMOTE_ADDR directly, but not the packed inet_pton values. Also we have the binary conversion in the install script.

svivian:
> If it is this, we may check if(isset() in the code before running qa_block_ip_match and return nothing if it is not set to prevent this.

The problem for `inet_ntop` is not just NULLs, it's any bad data, for example `inet_ntop('asd')`. As I said, this also includes the old integer IPs prior to the DB upgrade.

> Or we add an additional check something like qa_is_valid_ip which just test if if (filter_var($ip, FILTER_VALIDATE_IP)) returns true

That won't work for `inet_ntop`, since `filter_var` requires the IP string, not the binary. And to get the IP string we need to use `inet_ntop`... However, that *will* work for `inet_pton` - in hindsight I should have done that, I just went ahead and added @ without thinking :-/

And I totally agree about error suppression - actually for the most part I've been trying to remove error suppression wherever possible.

DanielRuf:
Right. Not so simple at first as I thought. Seems exepction handling would be useful here sometime in the future.

svivian:
I've just added some unit tests for the `qa_block_ip_match` function and there are a few cases that are failing:

	qa_block_ip_match('0:0:0:0:0:ffff:7f00:0001', '::ffff:7f00:*');
	qa_block_ip_match('::ffff:7f00:0001', '::ffff:7f00:*');
	qa_block_ip_match('::ffff:7f00:0001', '::*:7f00:*');

These all return false when they should be returning true.

DanielRuf:
I think the problem is that you can not use the compacted version in the second parameter. I might have mentioned this before, have to search for this.

DanielRuf:
The regular expression is ok, see regexpal.com. The IP does not match the range in the second parameter. You can see this when you output the IP and range variable.

svivian:
Yeah I guess that particular example doesn't make sense. Two placeholders at the end does, though. For IPv4 we can have `88.77.*.*` so for IPv6 it makes sense to have e.g. `::ffff:*:*`

Taking a look into this now. The problem comes from the `ipv6_expand` function.

```php
ipv6_expand('::ffff:7f00:*');
// result: '0000:0000:0000:0000:*:ffff:7f00:0000'
// expected: '0000:0000:0000:0000:0000:ffff:7f00:*'

ipv6_expand('::ffff:*:*');
// result: '0000:0000:0000:*:*:ffff:0000:0000'
// expected: '0000:0000:0000:0000:0000:ffff:*:*'
```

Looks like the wildcard replacement needs to work from the end. It only makes sense for wildcards to be at the end.

DanielRuf:
qa_block_ip_match('::ffff:7f00:0001', '::*:7f00:0001')

results in

IP
 string(39) "0000:0000:0000:0000:0000:ffff:7f00:0001"

range
string(36) "0000:0000:*:0000:0000:0000:7f00:0001"

the asterisk is at the wrong position

DanielRuf:
> Looks like the wildcard replacement needs to work from the end. It only makes sense for wildcards to be at the end.

Right, currently testing here. A safe approach would be to use the full expanded string in the second parameter.

DanielRuf:
added parantheses for the regular expression and counting index from the end.

http://ideone.com/OvUCbc

But generally it would be easier to use the full expanded string as it has 8 packets and is better to read.

The new code should fix these 3 examples and the others should still work. Currently 2am here, so I am not so good at thinking now ;-)

svivian:
Heh, I just came up with the exact same solution in my tests. Pushing a commit now.

What were the parentheses for in the regex? The tests all pass without those.

DanielRuf:
The parentheses should be ideally always there for grouping matches. It is just there to prevent potential issues with the regular expression. It is better all corresponding regular expressions have the same format like 0-9a-fA-F and the parentheses =)

Changed a line where A-f0-9 was written, please change it to 0-9a-fA-F

http://ideone.com/OvUCbc

svivian:
OK just pushed some more commits. I also found an issue with large numbers in the `ipv6_numeric` function (accuracy of `base_convert`). So I replaced it with something simpler that compares the array we get from `unpack`. Let me know if you spot any problems.

DanielRuf:
> So I replaced it with something simpler that compares the array we get from unpack.

Makes more sense comparing each single byte instead of the large single number which may take much longer and produce problems on some systems. So far had no issues with the code from before but it is better when we compary it byte by byte. :+1: 

And still works like expected http://ideone.com/NQWlay

svivian:
Our unit tests are run on PHP 5.3-7.0 via Travis: https://travis-ci.org/q2a/question2answer

I also have a Vagrant box with 5.2 that I check occasionally.

Issue #339, opened 2015-12-30
CSS Bug: In mobile view question title comes down counters if there are too many views


gurjyot:
Question title in questions list page shifts down in mobile if there are too many question views(I think this is the cause of the problem). See the below pic.

![q2a](https://cloud.githubusercontent.com/assets/6462850/12057736/0168dac4-af6a-11e5-8109-a8cf987ab0f7.png)

svivian:
Made a quick fix for this. But I'll have to take a closer look at the layout some time, because there are a few weird things with the size of the title block (`.qa-q-item-main`), sometimes it isn't filling the whole width.

PR #381, opened 2016-02-10
Vote box gap fixed 

### LINKED ISSUE #378: Vote box gap is showing on some pages after Q2A 1.7.3 update
On pages where vote box is not shown, there a gap equal to vote box is showing between answer count box and the question. This has happened after Q2A 1.7.3 update, before this update it was just perfect. 

test

amiyasahu:
Fixes #378

### DIFF
--- a/qa-theme/SnowFlat/qa-styles.css
+++ b/qa-theme/SnowFlat/qa-styles.css
@@ -1016,12 +1016,12 @@ blockquote p {
 }
 @media (min-width: 241px) and (max-width: 799px) {
 	.qa-q-item-stats {
-		width: 60px;
+		max-width: 60px;
 	}
 }
 @media (min-width: 800px) {
 	.qa-q-item-stats {
-		width: 133px;
+		max-width: 133px;
 	}
 }
 


svivian:
Unfortunately this doesn't quite solve the issue for me. The issue is occurring due to the fix for #339 where long view counts messed up the layout. With your fix here the layout gets misaligned again:
 
q2a-381

I think we can solve this by setting a different width on that page. Something like

    .qa-template-user-activity .qa-q-item-stats { width: 65px }

Is it just the user activity page the issue occurs on? Or are there any others?
/cc @gurjyot

amiyasahu:
@svivian , Sorry I missed this scenario earlier. 

So can we make the view count shorter. If we show like this and when the website gets older and have millions of views then we dont have to change the CSS again.

So I think it would be a wise idea to make the view shorter and make it human readable format like 5k views , 1M views something like that. 

I can make a PR for this if you agree.

svivian:
> So I think it would be a wise idea to make the view shorter and make it human readable format like 5k views , 1M views something like that.

> I can make a PR for this if you agree.

We already have an option for this in Q2A (thanks to pupi1985) which is part of v1.8. But not everyone will have that option turned on.

PR #382, opened 2016-02-11
Log any email sending errors

JaggedJax:
This can be very helpful when trying to figure out why emails aren't being sent.

### DIFF
--- a/qa-include/app/emails.php
+++ b/qa-include/app/emails.php
@@ -152,7 +152,11 @@
 			}
 		}
 
-		return $mailer->Send();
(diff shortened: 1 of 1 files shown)


JaggedJax:
Dang it. I meant to do this as two separate pull requests. Let me know if I should break this up.

The second commit disables SSL checks only if not using smtp_secure. I've had to do this in my environment because even though I request non secure and even though the smtp server has a valid certificate, I get an error that the CN record doesn't match. It's comparing the domain name to the ip address, so of course it's failing.

svivian:
Thanks for the PR!

PR #388, opened 2016-02-19
Support wordpress host port

dreamon11:
Support for WordPress MySQL alternate port : https://codex.wordpress.org/Editing_wp-config.php#MySQL_Alternate_Port


svivian:
Thanks!

PR #393, opened 2016-03-10
Out-of-the-box Joomla! integration.

Spudley:
These changes implement out-of-the-box integration with Joomla!, in a similar vein to the existing Wordpress integration.

In addition to these changes, the Joomla! integration also requires a plugin to be installed into Joomla! (see https://github.com/SFWLtd/plg_q2a_integration). This plugin allows the Joomla! administrator to configure the Joomla! permissions that will grant different levels of access to Q2A.

Note that the installation instructions on the Joomla! plugin currently describe how to install it with a version of Q2A that does not contain the changes in this pull request. The changes here will allow those installation instructions to be dramatically simplified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Q2A is highly customisable with many awesome features:
 - Private messages and public wall posts.
 - Log in via Facebook or others (using plugins).
 - Out-of-the-box WordPress 3+ integration.
+- Out-of-the-box Joomla! 3.0+ integration (in conjunction with a Joomla! extension).
 - Custom single sign-on support for other sites.
 - PHP/MySQL scalable to millions of users and posts.
 - Safe from XSS, CSRF and SQL injection attacks.
--- a/qa-config-example.php
+++ b/qa-config-example.php
@@ -128,6 +128,17 @@
 */
 
 /*
+	Out-of-the-box Joomla! 3.x integration - to integrate with your Joomla! site, define
+	QA_JOOMLA_INTEGRATE_PATH. as the full path to the Joomla! directory. If your Q2A
+    site is a subdirectory of your main Joomla site (recommended), you can specify
+    dirname(__DIR__) rather than the full path.
+	With this set, you do not need to set the QA_MYSQL_* constants above since these
+	will be taken from Joomla automatically. See online documentation for more details.
+
+	define('QA_JOOMLA_INTEGRATE_PATH', dirname(__DIR__));
+*/
+
+/*
 	Some settings to help optimize your Question2Answer site's performance.
 
 	If QA_HTML_COMPRESSION is true, HTML web pages will be output using Gzip compression, if
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -56,11 +56,16 @@
 
 	//	If we're using single sign-on integration (WordPress or otherwise), load PHP file for that
 
-		if (defined('QA_FINAL_WORDPRESS_INTEGRATE_PATH'))
+		if (defined('QA_FINAL_WORDPRESS_INTEGRATE_PATH')) {
 			require_once QA_INCLUDE_DIR.'util/external-users-wp.php';
-		else
+		}
+		elseif (defined('QA_FINAL_JOOMLA_INTEGRATE_PATH')) {
+            //grab the integration file that is included as part of the Joomla plugin (we already confirmed that it exists).
+			require_once QA_JOOMLA_PLUGIN_SSO_FILE;
+		}
+		else {
 			require_once QA_EXTERNAL_DIR.'qa-external-users.php';
-
+        }
 
 	//	Access functions for user information
 
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -60,8 +60,7 @@
 			require_once QA_INCLUDE_DIR.'util/external-users-wp.php';
 		}
 		elseif (defined('QA_FINAL_JOOMLA_INTEGRATE_PATH')) {
-            //grab the integration file that is included as part of the Joomla plugin (we already confirmed that it exists).
-			require_once QA_JOOMLA_PLUGIN_SSO_FILE;
+			require_once QA_INCLUDE_DIR.'util/external-users-joomla.php';
 		}
 		else {
 			require_once QA_EXTERNAL_DIR.'qa-external-users.php';
--- a/qa-include/qa-base.php
+++ b/qa-include/qa-base.php
@@ -49,8 +49,14 @@
 	qa_initialize_php();
 	qa_initialize_constants_1();
 
-	if (defined('QA_WORDPRESS_LOAD_FILE')) // if relevant, load WordPress integration in global scope
+	if (defined('QA_WORDPRESS_LOAD_FILE')) {
+		// if relevant, load WordPress integration in global scope
 		require_once QA_WORDPRESS_LOAD_FILE;
+	}
+	elseif (defined('QA_JOOMLA_LOAD_FILE')) {
+		// if relevant, load Joomla JConfig class into global scope
+		require_once QA_JOOMLA_LOAD_FILE;
+	}
 
 	qa_initialize_constants_2();
 	qa_initialize_modularity();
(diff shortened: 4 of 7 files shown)


svivian:
Hi! Firstly, apologies for not responding to your email or your posts on the Q2A site, I've been quite busy lately.

Thanks for taking the time to create a PR. However you don't have the external users integration file in your PR - looks like you import it from your plugin, is that right?

Is the Joomla plugin necessary for integration? The SSO functions in `qa-external-users.php` should cover everything you need for basic integration, including user levels (see `util/external-users-wp.php` for how we do this for Wordpress).

I was expecting to have a `util/external-users-joomla.php` file plus a few small clauses in the core.

---

As a side note, I've actually been considering _reducing_ the amount of "special-casing" we do for Wordpress, perhaps adding some extra SSO functions that any external user integration can define. For example instead of the WP code inside `qa_initialize_constants_1`, each SSO could have a `qa_external_initialize` function that we call. Then we just include `util/external-users-wp.php` and nothing more.

Spudley:
Hi Scott.

Thanks for the reply.

The Joomla integration I’ve written is much more two-sided than your existing Wordpress integration.

The main reason for this is that Joomla has a much more sophisticated user permissions model than Wordpress, and it needed a plugin on the Joomla side to provide a user interface for the site admin to configure how the Joomla access levels should map to the access levels in Q2A.

It is possible to integrate Q2A with Joomla at a basic level without any of this code; I had already previously posted an example qa-exernal-users script on the Q&A board on the Q2A site. However this example did not do a very good job of handling the permissions; I linked it with the core admin permissions built into Joomla, but that’s far from ideal and didn’t leave any flexibility for choosing who could be a moderator, etc on the Q2A side.

You are correct that in my PR I as importing the external-users file from the Joomla plugin. This isn’t strictly necessary; the reason I’ve done it that way is because I was already shipping my Joomla plugin prior to writing the code on the Q2A side, and I was shipping the external-users file with the rest of the code in the Joomla plugin. It was purely a pragmatic decision because I needed to ship it and it was convenient to ship it all in the same package. The external-users file I am shipping with the Joomla plugin is exactly the same as it would be if it was in the normal place except for a directory path when it loads the main include file from Joomla. It could just as easily have been in the code I submitted in the PR, and it probably would have been if I hadn’t already shipped the plugin.

As for the rest of the PR, as you’ll see, I’ve basically cloned as much as I could from your Wordpress integration. I already had good integration working via the Joomla plugin and the external-users file, so writing the code for the PR was more about trying to reduce the work involved for installing, ie by allowing it to pick up Joomla’s database credentials, etc.

With the PR in place, there is obviously now quite a bit of code that is duplicated or very similar for WP and Joomla in the Q2A core. There would obviously be scope for refactoring all of that, and I fully agree that moving it out into the external-users file would be a good idea. I had considered doing something similar as part of the PR but I felt it would have been too big a set of changes in one go; I figured I’d get the PR accepted first and then see what refactoring we could do from there.

Hope that goes some way toward explaining my thought process with the PR. Happy to discuss further if you have any other questions.

svivian:
Is there a way to do this without requiring the Joomla plugin? For example in `qa_get_logged_in_user` check if the plugin is available and if not, revert back to a basic check that they're a Joomla admin.

Then it is true "out of the box" integration, and we can recommend your plugin for people who want more control.

Spudley:
Hi. Yes, that would be possible. It'd only be able to offer very basic permissions, but it would be doable. If you'd like me to make the changes to the PR to do that, I'd be happy to go ahead when I get some free time.

svivian:
Yes I think that would be best, thanks.

Spudley:
Hi Scott. Just to let you know, I have now completed an additional set of changes to the PR, so that the Joomla integration is no longer dependant on the Joomla plugin being installed. It does still use the plugin if it's there, and the plugin does provide better functionality so is still recommended, but you can now integrate them without using it.

As you'd expect, this change involved pulling in a qa-external-users file with the functionality that I originally wrote way back when I first posted about integrating them on Q2A's own Q&A board. I've refactored it significantly though, to allow it to read from the Joomla plugin if it exists before falling back to the defaults.

The copy of the qa-external-users file that is supplied with the plugin is no longer used. (I will leave it in the distribution for the plugin though, as it would be needed by anyone trying to integrate to Q2A 1.7)

Hope you're happy with the changes.

Kind regards,


   Simon.

svivian:
Thanks, this all looks good. I'll check it out and test in more detail next week. Which versions of Joomla are supported?

Spudley:
Hi. I've tested this against the current version of Joomla, 3.4.8.

I haven't tested it against anything older, but the Joomla APIs I've used are very stable (Joomla is very good at maintaining backward compatibility), so I would expect it to work for earlier versions, at least as far as 3.0, but I haven't tested and I can't guarantee it.

Also note that Joomla 3.5 is due out in the next few days; I haven't tested this yet either, but I didn't see anything that would be significant to this integration in the change log; I'd be amazed if there were any problems. (I will test it in due course though, just to be sure)

Spudley:
Hi. Just to say that I have now tested it against Joomla 3.5 as well. All looks good.

svivian:
Sorry for the delay, one of my clients had a big problem with their site when I was due to test this... working on it now.

svivian:
Also, do you have a link for the Joomla plugin? Would it be [the Github repo](https://github.com/SFWLtd/plg_q2a_integration) or is it on the Joomla Extensions site?

Spudley:
Hi Scott.

No worries about the delay -- paid work has to come first!  :)

Yes, you can use the github repo as linked, or it is on the Joomla Extensions site, here: http://extensions.joomla.org/extensions/extension/communication/question-a-answers/integration-for-question2answer

svivian:
I'm running into a few problems with this. Mostly I think it stems from guest users not being able to view anything - is that intentional? Anyone not logged in would normally be able to view the Q2A site.

The other problem comes from the redirects when not logged in. I set up my test Joomla install in a subfolder so with the above error I get redirected to `/` instead of `/joomla/`. The `onGetURLs` also uses `.html` links for login/logout which as far as I know are not standard. Seems like the default, in J3.5 at least, is `/index.php/login`

svivian:
I've removed the redirects, and set the login/register URLs without the `.html`. I wanted to use `JURI::root()` but that kept returning the path including the Q2A subdirectory, not the actual Joomla root. If you know how to fix that let me know :)

Spudley:
Hi Scott.

Thank you for doing that.

Before I start, I should say that I've also got some paid work that needs doing now -- I've got a big project starting on Monday with a tight deadline, so I may not be as responsive for the next couple of weeks.

Re guest users not being able to view anything -- hmm, not intentional, but that is how my client's site works. I was sure I'd added code to make it work for guests, and I was sure I'd tested it, but I can see how it might have slipped in. I'll double check.

Re default redirect URLs: One of the the first things I do when I install Joomla is switch on the SEO-friendly URL options, so the URLs become /login.html, etc. But you're right, the defaults are less friendly. There is a Joomla method `JRoute::_()` which will take an unfriendly 'default' URL and convert it to a friendly URL if it exists. See [this Joomla manual page](https://docs.joomla.org/Supporting_SEF_URLs_in_your_component) for more info on this, and [this answer on slashdot](http://stackoverflow.com/questions/17638877/how-to-get-joomla-login-form-url-for-redirection) for how to apply it to get the correct login URL regardless of how it's actually configured in Joomla.

The above ought to allow you to avoid using `JURI::root()` at all. If you do still need to us it, I don't know why it is giving the wrong URL. Without checking the Joomla code itself, my guess is that it takes its starting point from the current base PHP program, which would normally be Joomla's index.php, but in this case is actually Q2A's index.php. But that's just a guess.

Spudley:
Okay, I've modified the code to use `JRoute::_()` for the login/logout/register URLs. That should play nicer with Joomla however it's been set up.

Spudley:
I think the guest access thing is to do with line 145 in qa-joomla-helper.php, but I don't have time now to change it and test it, so I'll have to come back to it.

svivian:
Just to let you know, I've made a bunch of changes in my local branch that fixes most of the issues. Will push to the 1.8 branch when I get a moment.

Spudley:
Ah awesome. Thanks for that, and thanks for letting me know.  :)

svivian:
OK I've fixed all the issues I ran into, and merged into the 1.8 branch. Can you check it out when you have time and confirm it works for you?

I couldn't find a way to resolve the username issue [mentioned above](#issuecomment-207471699) so on Q2A it's showing the username on all posts, not the display name.

Thanks for all the work on this! Hopefully this can increase the popularity of Q2A :)

PR #396, opened 2016-03-13
Replace emoji chars to HTML entitles

phy25:
Not fully tested yet, but I think this is way better than just deleting those emoji chars.

### DIFF
--- a/qa-include/util/string.php
+++ b/qa-include/util/string.php
@@ -535,17 +535,21 @@
 	}
 
 	/**
-	 * Removes 4-byte Unicode characters (e.g. emoji) from a string due to missing support in MySQL < 5.5.3.
+	 * Replace 4-byte Unicode characters (e.g. emoji) with HTML entitles due to missing support in MySQL < 5.5.3.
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for the PR. I actually didn't convert the characters intentionally, for two reasons:

1. They are removed from usernames as well as posts, and usernames cannot use HTML entities (or rather, a username with emojis would end up like `username&#x12345;`
2. Posts can use other formats than HTML (text, Markdown, etc). So like usernames it's not appropriate to keep the entity in all cases.

Potentially we could detect if a post is using HTML and convert to entities in that case (using a separate function like `qa_convert_utf8mb4`). While removing them in all other cases. What do you think?

phy25:
I forgot the HTML limitations in q2a before. Your solution is fine, but as MySQL has eventually support UTF8mb4, and on mobile phones people tend to use emoji more often (they are characters in deed), full support will be better.

Maybe it's too heavy to introduce an existing library to handle emojis. I think we can transfrom emojis into decoded format (like the HTML entitles), and when displaying in the page, convert them back (into HTML or raw characters) through some function.

Maybe I should close this PR and turn it into a issue?

PR #400, opened 2016-03-29
Remove old code format, it is not necessary

fulldecent:


svivian:
Will merge this some time in the future. Don't want to invite style-only changes just yet to avoid any conflicts between the `dev` and `1.8` branches. I also have some tools that should be able to handle most of this.

PR #403, opened 2016-04-04
Japanese Language

naga0ka:
japanese lang


svivian:
Hi, thank you for making a language file. However we don't add them to the core of Q2A, they are kept separately and listed on our [Add-Ons page here](http://docs.question2answer.org/addons/).

We did have a Japanese language file but it's no longer available, so it would be great if you could make a separate repository with your files. I will then add a link from the docs.

naga0ka:
Hi,svivian
Thank you for your comment.
We've created a different repository, please check
bye

svivian:
Thank you for doing this. I've added a link to your translations at http://docs.question2answer.org/addons/

PR #406, opened 2016-04-09
New answer with Ajax inserted at the bottom

phy25:


DanielRuf:
Why bottom? It is just once and it makes sense to put it to the top (UX). insertBefore is right imho =)

phy25:
When there are already some answers, answer form is at the bottom, not the top. Thus if you finished your answer, you don't need to scroll to the top to see your answer.
(Hopes I haven't modified my theme :joy:)

DanielRuf:
But append adds it to the end of the selector. prepend(To) and insertBefore add it to the beginning.

phy25:
I don't think there is any problem with the code. If the strategy is not suitable, feel free to discuss.

DanielRuf:
Mh, your solution could make more sense as StackOverflow does it this way. The form and new answers are at the bottom. The later is already the case in question2answer but the form is still on the top, you are right. Would improve the UX and be more consistent.

svivian:
Sorry I meant to comment on this earlier. It looks like from the code that only the answer is added at the bottom but the answer form is still opened at the top?

I think that the submitted answer should be placed at the same point it's being written, but that could be the bottom. (On my site I actually do this using a theme override.)

In other words, we should add an option for the answer form location (Top/Bottom) then the answer stays at that location when submitted.

DanielRuf:
> bottom but the answer form is still opened at the top?

Right.

> In other words, we should add an option for the answer form location (Top/Bottom) then the answer stays at that location when submitted.

Great idea. This is also what I head in my head, form and answer at the same location which improves UX and usability and reflects how this is on SO and other platforms =)

phy25:
I see. So the Javascript code should detect the position of the form and insert the new answer into the proper position.

I will modify the pull request later this week.

svivian:
I *think* you don't need to modify the Javascript at all, just move the answer form location to be after the answers. When the answer is added it replaces the form with the answer content.

PR #407, opened 2016-04-09
Change from / reply-to for feedback form

phy25:
If the email address user inputed has forced SPF check, the email will not be received by major ESPs unless it is put in `Reply-to` field instead of `From` field.

### DIFF
--- a/qa-include/app/emails.php
+++ b/qa-include/app/emails.php
@@ -130,7 +130,7 @@
 		$mailer->From=$params['fromemail'];
 		$mailer->Sender=$params['fromemail'];
 		$mailer->FromName=$params['fromname'];
-		$mailer->AddAddress($params['toemail'], $params['toname']);
+		$mailer->addAddress($params['toemail'], $params['toname']);
 		if(!empty($params['replytoemail'])){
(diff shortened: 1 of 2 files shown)


qa-include/app/emails.php
```diff
@@ -131,6 +131,9 @@ function qa_send_email($params)
 		$mailer->Sender=$params['fromemail'];
 		$mailer->FromName=$params['fromname'];
 		$mailer->AddAddress($params['toemail'], $params['toname']);
+		if(!empty($params['replytoemail'])){
+			$mailer->addReplyTo($params['replytoemail'], $params['replytoname']);
+		}
```
DanielRuf: isn't it AddReplyTo (uppercase a)?

phy25: The older code is wrong, yes: https://github.com/q2a/question2answer/blob/dev/qa-include/vendor/PHPMailer/class.phpmailer.php
Maybe I should change all of them :joy:

svivian:
Looks like PHPMailer's docs say the functions are title case, but the actual code says otherwise. Functions are case-insensitive in PHP so no matter. Anyway thanks for the PR.

PR #409, opened 2016-04-14
captcha language option

q2apro:
added language option for captcha, for internationalization

### DIFF
--- a/qa-plugin/recaptcha-captcha/qa-recaptcha-captcha.php
+++ b/qa-plugin/recaptcha-captcha/qa-recaptcha-captcha.php
@@ -45,13 +45,19 @@ class qa_recaptcha_captcha
 		if (qa_clicked('recaptcha_save_button')) {
 			qa_opt('recaptcha_public_key', qa_post_text('recaptcha_public_key_field'));
 			qa_opt('recaptcha_private_key', qa_post_text('recaptcha_private_key_field'));
+			qa_opt('recaptcha_language', qa_post_text('recaptcha_language_field'));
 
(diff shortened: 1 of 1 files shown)


svivian:
Is there a need for this? reCAPTCHA should default to the language of the visitor automatically.

q2apro:
No, it does not default. Otherwise I had not implemented it...

svivian:
According to [this](https://developers.google.com/recaptcha/docs/display#config) it does. I tested and it's working for me (browser language set to French):

captcha lang

Maybe you don't have the browser language set?

q2apro:
I am always using an English browser but surfing German, French and Lithuanian websites. 

The browser language is not a guarantee that the user / website uses this language as well. Again one of those big fails in usability.

svivian:
Yes I suppose that users may have their browser set up incorrectly. Plus if a site is in German it's fair to assume any visitors registering/posting should be able to read German. Will test & merge when I get time.

DanielRuf:
Thought about this. Why not define a sitewide language (too) which is added to the whole site using `<html lang="...">`?

svivian:
We already have a site language option, so I've added this to the reCAPTCHA plugin instead of using a separate option, and also added it to the `<html>` tag. Thanks for bringing up the issue.

q2apro:
... as long as the language codes fit: https://developers.google.com/recaptcha/docs/language

svivian:
Yes they are all standard language codes. If the site is using a language that reCAPTCHA does not support, it will fall back to the user's browser setting again.

PR #411, opened 2016-04-27
Fix to prevent losing of content

q2apro:
Read problematic scenario at http://www.question2answer.org/qa/26327/ - confirmed by many of my angry users...

### DIFF
--- a/qa-include/ajax/answer.php
+++ b/qa-include/ajax/answer.php
@@ -36,7 +36,7 @@
 	);
 
 
-//	Check if the question exists, is not closed, and whether the user has permission to do this
+//	Check if the question exists, and whether the user has permission to do this
 
 	if ((@$question['basetype']=='Q') && !qa_user_post_permit_error('permit_post_a', $question, QA_LIMIT_ANSWERS)) {
 		require_once QA_INCLUDE_DIR.'app/captcha.php';
--- a/qa-include/ajax/answer.php
+++ b/qa-include/ajax/answer.php
@@ -38,7 +38,7 @@
 
 //	Check if the question exists, is not closed, and whether the user has permission to do this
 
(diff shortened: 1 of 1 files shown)


svivian:
This isn't the correct solution as it lets users post answers on *any* closed question, defeating the purpose of closing in the first place. I understand how this is a problem though. Ideally we'd want to do either (or both) of these:

(a) Use an AJAX check every few seconds to check if the question has been closed, and display a warning if it has. Stack Overflow does something similar. However this would add quite a bit of overhead.

(b) If a question is closed as a duplicate while an answer is being posted, redirect the user to the duplicate question, with the answer form filled out, so they can submit the answer for the other question if appropriate.

q2apro:
c) allow answers on closed questions that get converted to comments later on by the moderators or admin. Memo: The answer button will not appear frontend when the question got closed and the user loads the page again, that's why I see the counter argument "lets users post answers on any closed question" as void.

svivian:
Just because the answer button does not appear, doesn't mean answers cannot be posted. Anyone can POST any data to any URL on a website. So it does not stop anyone posting to a closed question even years later.

q2apro:
And you think this is a common case? I think the case that content gets lost, is much more often than the case that someone uses a script to post content to a closed question. Lost content hurts the user. A posted answer, by whom so ever, can be deleted, converted or whatsoever. 

When I only had 1000 users, this was already a big problem. Now I am on 5k users and always need to hack this line to prevent more complains. Gosh.

svivian:
It's not about whether it's a common case, it's about the fact it's *possible*. It's essentially a security problem as it leaves your site open for abuse. Sorry if I'm not explaining it very clearly.

As I already said I understand the problem, the options I posted above would solve the problem of lost content. I'll certainly look into it when I can.

PR #419, opened 2016-06-16
Making headline text accessible for CSS

q2apro:
If the text within the H1 does not have a span with class, it cannot be selected separately.

### DIFF
--- a/qa-include/qa-theme-base.php
+++ b/qa-include/qa-theme-base.php
@@ -723,11 +723,16 @@ class qa_html_theme_base
 		$url = isset($q_view['url']) ? $q_view['url'] : false;
 
 		if (isset($this->content['title'])) {
-			$this->output(
-				$url ? '<a href="'.$url.'">' : '',
-				$this->content['title'],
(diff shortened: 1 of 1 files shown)


svivian:
Can you explain the problem in more detail? I don't understand what this is solving.

q2apro:
In my case, I need to reposition the text which is not possible directly because I cannot access it directly. I can only move the entire H1 which also repositions the favorite star.

In my opinion, the favorite star should under no cirumstances part of the H1 (!)

svivian:
Yes I agree, having the star inside the `<h1>` tag isn't great. Perhaps wrapping the h1 in a div would work better? Something like

    <div class="qa-page-heading">
        <span class="qa-favoriting">...</span>
        <h1>...</h1>
    </div>

Now you can target the elements separately. What do you think?

q2apro:
Better than before :)

svivian:
Now implemented. Note: custom themes may need some minor tweaking, mainly if absolute positioning is used.

PR #439, opened 2016-08-13
Making category barrier line more visible. The previous colour is bar…

gurjyot:
…ely visible.


svivian:
This seems a bit too far the other way now (i.e. too dark). Probably something around `#e0e0e0` would be best. I think we'd also need to change the colour in a few other places like `.qa-nav-cat-item`

gurjyot:
Yeah that will also do but, slightly more dark would be better...

svivian:
OK went through the various elements with borders and darkened them all where necessary.

Issue #440, opened 2016-08-19
PHP Warning:  session_start(): Cannot send session cache limiter | asktitle.php:76


q2apro:
With all server setups (shared, managed, own) that I tested/tried over the last years together with q2a, I have got this error once in a while: 

```
PHP Warning:  session_start(): Cannot send session cache limiter - headers already sent (output started at /qa-include/ajax/asktitle.php:76) in /qa-include/app/users.php on line 146

PHP Warning:  Cannot modify header information - headers already sent by (output started at /qa-include/ajax/asktitle.php:76) in /qa-include/app/users.php on line 183 
```

Looking to the line 76 in asktitle.php: 

```
echo "QA_AJAX_RESPONSE\n1\n";
```

Line 146 of users.php: 

```
session_start();
```

Line 183 of users.php: 

```
setcookie('qa_session', $handle.'/'.$sessioncode.'/'.($remember ? 1 : 0), $remember ? (time()+2592000) : 0, '/', QA_COOKIE_DOMAIN);
```

Seems to be an ongoing bug, reported here as well (qa-include/ajax/asktitle.php:76): 

http://www.question2answer.org/qa/44404/
http://www.question2answer.org/qa/48672/
http://www.question2answer.org/qa/48414/
http://www.question2answer.org/qa/30001/
http://www.question2answer.org/qa/49565/

What could be the problem and what the solution? 

I guess something before `session_start` in `asktitle.php` should be the culprit.

Just another hint: The `session_start();` is within `function qa_start_session()` which gets called with `qa_set_logged_in_user(...)` and `function qa_get_logged_in_userid()`. So there might already be the problem. 

Maybe also `qa_session_var_suffix()` is one candidate.

**q2a version 1.7.4**

q2apro:
Fixed by calling: 

    $userid = qa_get_logged_in_userid();

In the beginning of `asktitle.php`

q2apro:
https://github.com/q2a/question2answer/pull/442

q2apro:
@svivian Why have you removed the fix again? 

https://github.com/q2a/question2answer/commit/b86d42567a7c526e6eb3d504141cbe7d11ba3537#diff-1b1cce71c594ecec322006da97fb8ebf

Is there another fix for the problem now?

svivian:
The problem was solved in a different way, see #443. So this fix was no longer necessary, and the $userid is not used in the script.

PR #442, opened 2016-08-26
Fix of issue 440

q2apro:
https://github.com/q2a/question2answer/issues/440


svivian:
Thanks, I'll check into this when I get a chance.

q2apro:
@svivian Why have you removed the fix again? 

https://github.com/q2a/question2answer/commit/b86d42567a7c526e6eb3d504141cbe7d11ba3537#diff-1b1cce71c594ecec322006da97fb8ebf

Is there another fix for the problem now?

Issue #447, opened 2016-09-08
External Login page error


jenolan:
When using external authorisation if the user loads http://www.question2answer.org/qa/login directly rather than use the external auth it loads the page which at the very top throws an error "PHP Question2Answer fatal error: User login is handled by external code" it should actually flip over to the external login!

Also, 'register' and 'logout' behave the same if the direct link is used it doesn't go to the external auth system.

jenolan:
Should it be something like this?
`//	Check we're not using Q2A's single-sign on integration and that we're not logged in

	if (qa_is_logged_in())
		qa_redirect('');
	if (QA_FINAL_EXTERNAL_USERS)
	{
		$userlinks=qa_get_login_links(qa_path_to_root(), isset($topath) ? $topath : qa_path($request, $_GET, ''));
		if (!empty($userlinks['login']))
    		qa_redirect_raw($userlinks['login']);
		qa_fatal_error('User login is handled by external code');
	}`

PR #448, opened 2016-09-09
Issue #447 Page error wit enternal auth

jenolan:
handle login, logout & register page calls when external  auth is enabled


svivian:
Thanks for this, I'll take a closer look as soon as I get the chance.

svivian:
Sorry for the delay. Thanks for this, it has been merged into the 1.8 branch.

PR #452, opened 2016-10-09
Fix double slash in default site_url option

pupi1985:
When `$_SERVER['SCRIPT_NAME']` is at the root such as `/index.php` then calling `qa_default_option('site_url')` generates a double slash. This commit removes it.


svivian:
Hi Gabriel, nice to see you active again :) I've been extremely busy the past little while but I'll merge this when I get a moment.

PR #454, opened 2016-10-18
Updated version.txt

ProThoughts:
Updated version.txt for v1.8


svivian:
Thanks for the PR, but I don't usually change the version number unless there's an official release. Because it can cause merge conflicts with other branches (e.g. if a 1.7.5 release happens, the `dev` branch will conflict with the `1.8` branch). Did you need this changed for a particular reason?

BTW when the version is updated we need to change the constant at the top of `qa-base.php` too. I'd also use something like `1.8-alpha` for the version as it's not the final 1.8 release.

ProThoughts:
Ok, got it. may be you should change to at least to 1.8-beta or 1.8-alpha so that people knows it is under testing.

PR #456, opened 2016-10-20
Add form file field to base theme

pupi1985:


svivian:
Good idea. I noticed you have a value attribute in there, but unless I'm mistaken the file input doesn't take a value, right?

pupi1985:
Oh, I thought it was sent somewhere in the request. I just took a quick look and couldn't find it anywhere so let's remove it.

PR #455, opened 2016-10-20
Improved Forgot Password Procedure

q2apro:
_Preamble: Did not know how I can prevent the other two commits from 2015..._

**NEW:**
1. Forgot password, enter username or email
2. Mail goes to user, front end shows: check your email
3. User clicks link in email, comes to reset page, data (code and email/handle) gets submitted by JS automatically
4. User gets the 2nd email that contains his new password, frontend shows: "password was sent to you", and the input fields, cursor in password field blinking
5. User checks his email, copies the password, done. 

Not perfect, but better than before!

PR #457, opened 2016-10-22
Remove unnecessary step in password reset

pupi1985:
I took some of the suggestions in #455 and tried to make the forgot password process a bit shorter. In this branch, the user should navigate to `/forgot` and input their username as they did before:

This will do 2 things: one would be to send the same first email as before (with a link to the reset password page containing the code and the shortcut link). The other would be to display a form to input only the code:

If the user inputs the code manually and clicks on the _Next step_ button, this will take them to the password confirmation page (which is still in the reset.php file)

If the user instead of manually writing the code, prefers to click the link in the email, then he will be sent to the password confirmation page directly.

Once the user inputs a valid password they will get automatically logged in. Alternative, the user could be sent to the login section, but this seems to be unnecessary, IMO.

In the previous process it was possible to reset the password from `/forgot` and from `/reset`. Now it is no longer possible to reset from `/reset` as the username is input in `/forgot`

### DIFF
--- a/qa-include/app/users-edit.php
+++ b/qa-include/app/users-edit.php
@@ -358,10 +358,12 @@
 	}
 
 
+/**
+ * Successfully finish the 'I forgot my password' process for $userid, sending new password
+ *
+ * @deprecated This function has been replaced by qa_finish_reset_user since Q2A 1.8
+ */
 	function qa_complete_reset_user($userid)
-/*
-	Successfully finish the 'I forgot my password' process for $userid, sending new password
-*/
 	{
 		if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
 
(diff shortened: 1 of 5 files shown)


yerbol89:
@pupi1985 this is nice core hack. But send Code: its not very nice. My site users its complicated.
Not send Code. Please send link and user open email and click link after show new pass edit page. 
https://cloud.githubusercontent.com/assets/1449790/19617486/46539da4-9808-11e6-867b-fafe1948657f.png

pupi1985:
Sorry, but I'm not quite sure I understand what you're asking for. Can you clarify your expected steps?
The process I've implemented is quite standard. In fact, try to reset your Facebook's password and it will follow exactly the same process.

amiyasahu:
@pupi1985, If I am not wrong he is trying to convey the below message. 
> You have done a pretty good job with this PR. But sending security code to the user and asking him to type it on the website is not a good idea. Rather we can send a link to the user via email which will contain the code and when the user clicks the link, it will bring him to the third page of password reset. (where he is just being asked to enter the new password). In this case, we can avoid second step.

I also think this is a good suggestion. What do you think?

pupi1985:
It makes sense. But that is exactly what I have implemented. That's why I didn't think yerbol was talking about that:

> If the user instead of manually writing the code, prefers to click the link in the email, then he will be sent to the password confirmation page directly.

amiyasahu:
Yes, you have already implemented that. 

So do we still need the second page for asking to enter the security code?

I think the second page can be **completed removed**, which makes the process more user-friendly.

pupi1985:
Maybe the email client the user is using blocks links so the user can't just click it. Copy/pasting a piece of text with the length of an URL, which might get a bit long is some times, and manually pasting is as a URL is likely to be harder than just copying one word and pasting it in the code input. E.G.: Whole process in Linux would take a double click to select and copy the code and a middle click to paste it. Pretty fast.

For more information, reset your Facebook password. If they do it this way and they're one of the most visited sites in the world it can't be that bad :)

svivian:
Sorry for the delay on this. All looks good from first glance, will test sometime this week.

svivian:
Thanks for this! Great work.

Jonatan442:
Is this change optional?
Some Q2A users probably prefer the old system to change passwords.

svivian:
No it's not optional, sorry. The old process was slow and insecure, the new method is much better and follows how pretty much every other site does it.

PR #459, opened 2016-10-24
IP v6 address handling

tonycleveland:
Changed datatypes and inet functions to handle ipv6 addresses on new installations

### DIFF
--- a/qa-include/db/admin.php
+++ b/qa-include/db/admin.php
@@ -159,7 +159,7 @@
 */
 	{
 		return qa_db_read_all_values(qa_db_query_sub(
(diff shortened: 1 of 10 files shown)


svivian:
Thanks for the effort, but unfortunately IPv6 was already been implemented a few months ago (see the `1.8` branch). Partly my fault for not having released v1.8 sooner I guess. BTW as @pupi1985 said, those MySQL functions are 5.6 only so we had to use the PHP functions.

In future, just double check with us here or on the [Q2A site](http://www.question2answer.org/qa/) so you don't waste any effort :)

PR #461, opened 2016-11-06
Fetch latest released Q2A version from GitHub itself

pupi1985:
I know this might be a kind of political commit but it:

- Improves the look and feel of the number as the links are now rendered by the theme and not in an iframe
- Follows the `all development is taking place through GitHub` premise in the README
- Saves time when releasing a new Q2A version
- Decouples the need of getting both links updated before releasing which means more independance
- Slightly increases privacy :)

### DIFF
--- a/qa-content/qa-admin.js
+++ b/qa-content/qa-admin.js
@@ -134,9 +134,9 @@ function qa_admin_click(target)
 	return false;
 }
 
-function qa_version_check(uri, version, elem)
+function qa_version_check(uri, version, elem, isCore)
 {
-	var params={uri:uri, version:version};
+	var params={uri:uri, version:version, isCore:isCore};
 
 	qa_ajax_post('version', params,
 		function (lines) {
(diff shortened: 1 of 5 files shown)


pupi1985:
Good catch

svivian:
I'm wondering if this should be an AJAX call like we have with the plugins. If GitHub goes down (which is rare, but they do get targeted a lot) it will slow down/stop the page loading.

pupi1985:
Yes AJAX would be a better choice. Only thing is that it will require some more additional changes as the core would be fetching its new version differently from plugins (which use the metadata). So that would mean making a small tweak to the JS performing the check and a few more changes to the ajax/version.php file.

I've just rebased the branch to contemplate these changes.

Alternatively, for faking the default parameter in JS < ES6:
`isCore = typeof isCore !== 'undefined' ? isCore : false;`

svivian:
I've just noticed that the current (iframe) version check has the language string is passed into it. As far as I can see that has no effect but I'll double check with Gideon.

Ping @gidgreen

pupi1985:
Yes, that's why I said `Slightly increases privacy` :P

gidgreen:
Sure, this makes sense, and gives me one less step to take care of for each release :)

The `version` and `language` parameters were originally put in there so I could collect some empirical data on usage patterns – obviously no issue removing them.

PR #464, opened 2016-11-08
A few filter-basic fixes

### LINKED ISSUE #422: Comments use answer length limit
As reported [here](http://www.question2answer.org/qa/52227/comment-character-limit-using-answer-character-limit-normal). When setting different comment and answer length limits, Q2A uses the answer limit for comments.

pupi1985:
Should fix #422


svivian:
Thought I'd already fixed this but maybe not. Thanks for the PR.

PR #467, opened 2016-11-09
Add permission to define what users can see the special users page

pupi1985:
Note moderators+ will still be able to see the blocked users.

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -1299,26 +1299,37 @@
 	Return the sub navigation structure for user listing pages
 */
 	{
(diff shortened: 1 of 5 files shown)


svivian:
Another great PR. I've made a few changes/additions:

- Changed the URL to `users/new` as it felt more natural to me.
- Added permission for viewing the new users page as well. I think one of the main uses for the page is checking for spam accounts, so it's probably best the page isn't visible to everyone by default. Don't want a "shop window" for users/search engines to see spam accounts.
- Shortened the classes, for example `.qa-special-users-label` rather than `.qa-top-special-users-label`.

The one thing that's missing is adding those classes to the other themes. However, the CSS that's now in the SnowFlat theme is getting unwieldy. The only difference between the sections is the bag icon, I feel like there should be a better way but I can't figure it out right now.

PR #468, opened 2016-11-11
Improve avatar handling API

pupi1985:
I'm in a hurry so I haven't tested this one much.

  * Created function `qa_get_avatar_blob_url` that returns a link to a locally stored avatar
  * Created function `qa_get_gravatar_url` that returns a link to a Gravatar
  * Created function `qa_get_user_avatar_source` that returns the source that will be used for the given parameters
  * Created function `qa_get_user_avatar_url` that returns a link to an avatar, regardless of the source (this is actually the motivation for the PR)

Finally
  * Updated `qa_get_avatar_blob_html` to call `qa_get_avatar_blob_url`
  * Updated `qa_get_gravatar_html` to call `qa_get_gravatar_url`

Ref: http://www.question2answer.org/qa/32982

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -2042,7 +2042,7 @@
 	}
 
 
-	function qa_get_avatar_blob_html($blobid, $width, $height, $size, $padding=false)
+	function qa_get_avatar_blob_html($blobId, $width, $height, $size, $padding = false)
 /*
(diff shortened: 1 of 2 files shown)


svivian:
Thanks, good stuff. If you get a chance, it would be useful to list some of these on the functions page of the docs.

PR #469, opened 2016-11-12
Fix users being unable to see all their own profile fields

pupi1985:
1. Create a field only visible by moderators.
 1. Login as a registered user
 1. In the profile section you can not see the field
 1. In the account section you can see and edit the field

It is not clear if a user is supposed to see their own fields. However, the note says *(only shown to moderators)* which is also present in the email so by analogy a user can see all their own fields. It would also make sense when registering.


svivian:
Yes that makes sense. Perhaps we should change the message to something like "shown only to you and moderators"?

pupi1985:
Yes, that would make things clearer. These are the relevant strings:
```
		'only_shown_admins' => '(only shown to admins)',
		'only_shown_editors' => '(only shown to editors and above)',
		'only_shown_experts' => '(only shown to experts and above)',
		'only_shown_moderators' => '(only shown to moderators and admins)',
```
They're only used in the profile page so I guess changing that shouldn't have any negative impact. Maybe some plugin is using them... however, if a user is able to see it then it should be a user with a level higher then just a registered user so in the worst case scenario it will be slightly redundant to see "you and admins".

PR #471, opened 2016-11-25
Include confirmation code in confirmation email

pupi1985:
Just another awkward commit.

Imagine a user is using a Q2A mobile app. They request to change their e-mail address and the site requires e-mail confirmation. They open their mobile e-mail client. Tthe email they receive has a link which makes them open a 3rd app: the browser. The idea of the mobile app is not to depend on the browser but unless the user selectively copies the code from the URL they have no other option than opening the browser.

This way there is no dependency on the browser and it would be possible for the mobile app to ask the user to input the code manually, which can be easily copy/pasted.

### DIFF
--- a/qa-include/app/users-edit.php
+++ b/qa-include/app/users-edit.php
@@ -250,23 +250,29 @@
 
 		$userinfo=qa_db_select_with_pending(qa_db_user_account_selectspec($userid, true));
 
+		$emailcode = qa_db_user_rand_emailcode();
+
 		if (!qa_send_notification($userid, $userinfo['email'], $userinfo['handle'], qa_lang('emails/confirm_subject'), qa_lang('emails/confirm_body'), array(
-			'^url' => qa_get_new_confirm_url($userid, $userinfo['handle']),
+			'^url' => qa_get_new_confirm_url($userid, $userinfo['handle'], $emailcode),
+			'^code' => $emailcode,
 		)))
 			qa_fatal_error('Could not send email confirmation');
 	}
 
 
-	function qa_get_new_confirm_url($userid, $handle)
+	function qa_get_new_confirm_url($userid, $handle, $emailcode = null)
 /*
-	Set a new email confirmation code for the user and return the corresponding link
+	Set a new email confirmation code for the user and return the corresponding link. If the email code is also sent then that value
+	is used. Otherwise, a new email code is generated
 */
 	{
 		if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
 
 		require_once QA_INCLUDE_DIR.'db/users.php';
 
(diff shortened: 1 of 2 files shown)


svivian:
Thanks, this is a useful addition. However if we're listing the conf code separately, it would also be useful to allow the user to input the code on the website, same as we do for resetting password. What do you think?

pupi1985:
I think... it is much complex than it seems :) The reset password form is actually only one form displayed in two steps (a kind of wizard). Adding that to the confirm will make it a bit more complex as there is already a form for the Resend Password feature. Furthermore, in this case, it is possible for the user to be logged in and that is a variable that affects many parts of the business logic.

Assuming the idea is to be able to answer to these different scenarios (note each scenario might be different depending on whether the user is logged in or not):
 1.  `GET /confirm` to allow the user to input user/code and re-send code
 1.  `GET /confirm?u=X` to allow the user to only input the code and re-send code
 1.  `GET /confirm?c=X` to allow the user to only input the user and re-send code
 1.  `POST /confirm (resend-confirmation)`  to actually send the confirmation
 1.  `POST /confirm (confirm)` to actually perform the confirmation by submitting the form
 1.  `GET /confirm?u=X&c=Y` to try to perform the confirmation and display errors in the form if it fails and re-send code

Then it is likely that the level of statement nesting will reach 6 or 7. Not being able to reuse code by means of methods/functions will make it awfully long and hard to read. Not to mention the awkward situations generated by a logged in user that tries tu confirm a different handle in the URL (or submitting the form).

Getting to this page (same as the reset page) without a confirmation code and a valid user is unlikely and can be workaround by resending a code Now, assuming the code input is needed, what will be the minimal scenarios to tackle?

svivian:
Not sure I'm following you. All we'd need to do is add one text box to the view that appears after sending the code. That would only add one extra step in the code, to handle the POSTing of a code.

Merging in your current patch and will take a look at this tonight.

pupi1985:
OK. That is considerably simpler. That process would only apply to logged in users and the form should only be displayed after sending the confirmation link (it is not possible to access that form directly with a `GET` request).

I'm adding some commits to this branch. Note the current confirmation process has been reused: only one attempt allowed and, if failed, a new email has to be sent.

Maybe these strings should be renamed to be reused between the confirm and reset pages:
```
'reset_code_another' => 'send another',
'reset_code_emailed' => 'You have been emailed your code',
'reset_code_label' => 'Code:',
'reset_code_wrong' => 'Code not correct',
```

Such as:
```
'email_code_another' => 'send another',
'email_code_emailed' => 'You have been emailed your code',
'email_code_label' => 'Code:',
'email_code_wrong' => 'Code not correct',
```

Also feel free to rephrase any of the strings I changed. I'm not happy with `confirm_emailed` but I couldn't keep it shorter than that.

PR #473, opened 2016-11-29
Let user input a confirmation code manually

pupi1985:
Continuing #471


svivian:
Thanks. I went ahead with your suggestion to rename the language keys from `reset_code_*` to `email_code_*`. The `confirm_emailed` change didn't sound quite right to me so I've made it similar to how it was before. It's not too long IMO.

PR #477, opened 2016-12-18
Fix overrides not being called after enabling/disabling plugins PR merge

pupi1985:
My previous pull request about disabling plugins generated an issue. See the previous PR diff:
```diff
        qa_initialize_constants_2();
        qa_initialize_modularity();
        qa_register_core_modules();
-       qa_load_plugin_files();
        qa_load_override_files();
 
        require_once QA_INCLUDE_DIR.'qa-db.php';
 
        qa_db_allow_connect();
 
+       qa_load_plugin_files();
```

Plugin overrides get loaded from plugin files. So if we try to load them before the plugin files are loaded then the overrides will be empty. So they need to move that line after the `qa_load_plugin_files();`.

However, there is a silly issue. What about the `qa_db_allow_connect()` function? It seems it is a barrier to make sure nothing accesses the database before that point. The thing is, it can be overridden. So loading the overrides AFTER running that function will result in no effect from that specific override override. Making the `qa_db_allow_connect()` function not overrideable makes it almost useless.

Furthermore, the `qa_db_allow_connect()` doesn't actually provide any kind of additional security, as far as I see, because the plugins could still connect to the database without using the Q2A DB functions (they have access to the connection user/pass/host).

My suggestions/comments are:

 * Accept the fact the core needs to access the database before loading the plugin files
 * Remove the `qa_db_allow_connect()` function as there is no real benefit from it
 * To workaround the missing function it is still possible to override the `qa_db_connect()`
 * I hardly believe someone out there has actually overridden `qa_db_allow_connect()` (ever) so adding it to the changelog I guess it should be more than enough

**Edit: Another issue was found so also added the `Fix database fail handlers` commit**

Accessing the database before loading the plugin files really messed things up. Installation is not currently working. The DB fail handler that catches the errors and detects no table exist is not being assigned on time. I guess this should be enough to get you started.

Now, I found no simple and easy way to fix this without considerably rearranging code. This is what I did for this suggestion:

 * I split `qa-base.php` so that calling it doesn't force a database read (which was the change added when enabling/disabling plugins
 * `require 'qa-base.php';` should now always be followed by `qa_initialize()` in order to connect to the database and load plugin files
 * Had to slightly change the routing mechanism (nothing too deep)
 * Had to remove qa_report_process_stage('init_install')

Again, this is just a suggestion. I don't like it but I couldn't come with something that perform less radical changes to the core. I hope this at least servers to point at some parts of code that might need to be updated.point some parts of code.

### DIFF
--- a/qa-include/qa-ajax.php
+++ b/qa-include/qa-ajax.php
@@ -20,30 +20,6 @@
 	More about this license: http://www.question2answer.org/license.php
 */
 
-// Output this header as early as possible
-
-header('Content-Type: text/plain; charset=utf-8');
-
-
-// Ensure no PHP errors are shown in the Ajax response
-
-@ini_set('display_errors', 0);
-
-
-// Load the Q2A base file which sets up a bunch of crucial functions
-
-require 'qa-base.php';
-
-qa_report_process_stage('init_ajax');
-
-
-// Get general Ajax parameters from the POST payload, and clear $_GET
-
-qa_set_request(qa_post_text('qa_request'), qa_post_text('qa_root'));
-
-$_GET = array(); // for qa_self_html()
-
-
 // Database failure handler
 
 function qa_ajax_db_fail_handler()
@@ -52,6 +28,29 @@ function qa_ajax_db_fail_handler()
 	qa_exit('error');
 }
 
+// Ensure no PHP errors are shown in the Ajax response
+
+@ini_set('display_errors', 0);
+
+// Output this header as early as possible
+
+header('Content-Type: text/plain; charset=utf-8');
+
+// Load the Q2A base file which sets up a bunch of crucial functions
+
+require 'qa-base.php';
+
+qa_db_set_fail_handler('qa_ajax_db_fail_handler');
+
+qa_initialize();
+
+qa_report_process_stage('init_ajax');
+
+// Get general Ajax parameters from the POST payload, and clear $_GET
+
+qa_set_request(qa_post_text('qa_request'), qa_post_text('qa_root'));
+
+$_GET = array(); // for qa_self_html()
 
 // Perform the appropriate Ajax operation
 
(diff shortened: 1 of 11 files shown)


svivian:
This is an interesting problem. First, regarding the `qa_db_allow_connect` function: yes it could be bypassed, but that's a "hack" that would be discouraged. (Same as any developer could go in and remove the password check for logging in, doesn't mean it's a good idea...)

However, there is a reason for the function to exist - plugins may need to override database functions before those functions are called. For example if someone wanted to use Postgres instead of MySQL, or add logging to all queries.

We're in a catch-22 here: we can't check whether plugins are enabled until after we connect to the database, but we can't connect to the database until after override plugins are registered.

I can't think of any way forward at the moment. I don't think there is a sane way to check if a plugin will override the database, and make those always enabled.

pupi1985:
OK, you totally got the issue.

I also considered rolling it back too. Another alternative would be to load/save them from a (JSON?) file. The thing is that the file will need to be writable which I think complicates things a little bit for newcomers.

Another alternative that I'm brainstorming while writing it right now is that maybe we don't actually need to choose from overrides loaded after accessing the DB or before as maybe we can have both types of overrides. The only difference between them would be the point in the bootstrap process they are called at.

Finally, there is also the cheap approach (which is kind of how the external users is being handled) which could be importing a given file from a given directory (e.g.: `qa-include/plugins/db-overrides.php`) and document it so that the user is aware of this. It could be loaded while loading the system modules.

Anyway, rolling this back is a good option and most likely necessary in every alternative I thought of.

svivian:
Just discovered that this also broke the unit tests, so I've made a quick fix to disable the feature (076a40af9dec434c823fd85644539386c18967ab). Most of the code is still there, plugins can still be "enabled" and "disabled" in admin, but the plugin manager now loads all plugins as was the old behaviour.

> maybe we don't actually need to choose from overrides loaded after accessing the DB or before as maybe we can have both types of overrides

Not a bad idea. Perhaps we could use the plugin metadata to determine if a plugin must be run before the database is available.

pupi1985:
Good. I've made a new suggestion on PR #485 . Maybe it makes more sense to follow those changes there and close the PR

PR #485, opened 2017-01-08
Add ability to load plugins after and before connecting to the DB

pupi1985:
Summary of the changes
 * Split the `qa_load_plugin_files()` function into the parsing of the metadata files and the loading the plugins themselves
 * Moved the previous mentioned functions to the `PluginManager`
 * Modified how overrides are loaded: instead of being added to the global `$qa_override_files` variable they are queued in `$qa_override_files_temp` as a delayed load
 * Function `qa_load_override_files()` instead of loading all overrides from the global variable `$qa_override_files` it loads them from the temp one and also adds them to the original `$a_override_files` global variable just to keep backwards compatibility (probably unnecessary)
 * The plugin load (including loading the overrides) happen before and after connecting to the database
 * Added `load_order` to the `metadata.json` (it a string just to make sure we have more values to set than just a boolean in the future). Current values could be `after_db_init` and `before_db_init`.
 * Plugins that are loaded after the database is loaded can not be disabled
 * Default behavior for `metadata.json` files without `load_order` was set to load before. That way only plugins that specifically update their metadata for v1.8 will actually be able to be disabled. The opposite behavior could also be fine, however, it will bring issues to users that have overridden the core database functions

### DIFF
--- a/qa-include/Q2A/Plugin/PluginManager.php
+++ b/qa-include/Q2A/Plugin/PluginManager.php
@@ -26,13 +26,102 @@ class Q2A_Plugin_PluginManager
 	const PLUGIN_DELIMITER = ';';
 	const OPT_ENABLED_PLUGINS = 'enabled_plugins';
 
+	private $loadBeforeDbInit = array();
+	private $loadAfterDbInit = array();
+
+	public function readAllPluginMetadatas()
+	{
+		$pluginDirectories = $this->getFilesystemPlugins(true);
+
+		foreach ($pluginDirectories as $pluginDirectory) {
+			$pluginFile = $pluginDirectory . 'qa-plugin.php';
+
+			if (!file_exists($pluginFile)) {
+				continue;
+			}
+
+			$metadataUtil = new Q2A_Util_Metadata();
+			$metadata = $metadataUtil->fetchFromAddonPath($pluginDirectory);
+			if (empty($metadata)) {
+				// limit plugin parsing to first 8kB
+				$contents = file_get_contents($pluginFile, false, null, -1, 8192);
+				$metadata = qa_addon_metadata($contents, 'Plugin', true);
+			}
+
+			// skip plugin which requires a later version of Q2A
+			if (isset($metadata['min_q2a']) && qa_qa_version_below($metadata['min_q2a'])) {
+				continue;
+			}
+			// skip plugin which requires a later version of PHP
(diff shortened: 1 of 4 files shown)


svivian:
Looks good from first glance, will test later this week. 

Only issue is the unit tests are failing (despite Travis reporting that everything is working for some reason). Probably we need to manually set up a "cache" of loaded plugins to avoid hitting the DB during tests.

pupi1985:
I understand what you mean about caching the enabled options. However, regarding the failing tests, I have to confess I'm on Travis side :)

I ran the tests before pushing and everything succeeded in my end:
```
phpunit.phar --bootstrap qa-tests/autoload.php qa-tests
PHPUnit 5.6.8 by Sebastian Bergmann and contributors.

..........................                                        26 / 26 (100%)

Time: 148 ms, Memory: 15.50MB

OK (26 tests, 116 assertions)
```

I must be missing something but I can't figure it out. What tests are failing in your end?

svivian:
It's working for you because you presumably have a database set up for your test site, so it's connecting to the DB fine on your machine.

You can see the issue in Travis here: https://travis-ci.org/q2a/question2answer/jobs/190014841
You can mimic the error yourself by putting an incorrect password in your config.

I'll take a look at this tonight.

pupi1985:
Got it. I thought they were totally unrelated issues. So it is just a matter of wrapping the database with the global `$qa_options_cache`. I've just committed and pushed the change.

svivian:
Thanks for this. I made a minor change to prevent disabling of plugins that don't have `load_order` set, that should avoid confusion :)

I've also set `load_order` for the core plugins, except for the WYSIWYG editor (which will cause problems if disabled) and logging plugin.

pupi1985:
Great. I've hadn't updated the metadata from the plugins because I wasn't sure you were actually going to like the strings I'd chosen :)

svivian:
The only thing I wasn't 100% sure on was whether to require the `load_order` key. Ideally, since most plugins don't need to override the DB, they can be `after_db_init` by default (and thus allowed to be disabled). But like you said, that could cause problems. Possibly we could switch it around for v2.0.

ProThoughts:
below issue is still present please look into this patch again.
http://www.question2answer.org/qa/55467/anyone-tried-fresh-installation-latest-version-github-error

PR #496, opened 2017-02-25
Some more minor changes

pupi1985:
**Add default value for permit field in userfields initialization**
For some reason it was left null.
Slightly refactored the code too (hopefully making it easier to read).

**Make selects react on changing events rather than click events**


svivian:
Did you test the JS for the selects? It looks like you're applying the change event to the options, not the select as a whole. Does that still work?

pupi1985:
Yes, it works. This is because the click/change event is not being applied to the options but to the select.  Specifically, it iterates over the ids of the "parent" selects.

svivian:
OK my bad. Thanks for the PR.

PR #505, opened 2017-03-23
qa-lang-admin.php translated into Gorgian

comcoma:
This is Georgian translation of qa-lang-admin.php by Giorgi Tsomaia.


svivian:
Closing for #508

PR #506, opened 2017-03-23
This is meta file to show language in admin panel

comcoma:
It shows how Georgian will be displayed in Q2A admin panel (language dropdown).


svivian:
Closing for #508

PR #507, opened 2017-03-23
qa-lang-emails.php translation into Georgian

comcoma:
This is Georgian version of qa-lang-emails.php file. Originally translated by Giorgi Tsomaia (me).


svivian:
Closing for #508

Issue #510, opened 2017-03-28
Facebook Login not working


cksagherian:
The problem is causing the facebook login app to successfully login, then redirect to a different page. This isn't working on the main q2a actual site either. Please help this is extremely frustrating. what did facebook change? it was working just a few days ago. 

I currently have it turned off. I've also tried using login master and the same thing happens with google and everyone else. Is this a google/facebook issue?

Thank you in adavnce.

cksagherian:
There also seems to be a refresh loop happening on the main site for some reason as well. I'm wondering if this is related. Thank you

iamdand:
Same thing here. Loops on and on.

valan:
this fix worked for me:
http://stackoverflow.com/questions/42994019/facebook-graph-api-not-work-from-2-2-to-2-3

iamdand:
Worked for me as well. Thanks for the tip, @valan !

cksagherian:
For those who are not code savvy, I've compiled the new php file that should replace.

/public_html/qa-plugin/facebook-login/base_facebook.php

Simply delete or rename original and replace with this new file. 

NOTE**** GitHub only allows .txt files to be uploaded so just rename to .php.

The credit for this fix should be given to @iamdand. I'm waiting to hear back on whether there are any security issues that would cause this to not be used on a "Production Site" or if he simply doesn't want blame for people copying and pasting the wrong code and therefore screwing something up. ;-)

I have verified this to work though.
 
[base_facebook.txt](https://github.com/q2a/question2answer/files/879387/base_facebook.txt)

ProThoughts:
Thank you all! submitted this patch here https://github.com/q2a/question2answer/pull/512

PR #512, opened 2017-04-01
Facebook login issue #510

ProThoughts:
@Scott, @pupi1985

This fix is written and verified by http://question2answer.org/qa/user/Dan+D. in below thread http://question2answer.org/qa/56966/facebook-login-stopped-working-its-not-even-working-your-site?show=56992#a56992
also see https://github.com/q2a/question2answer/issues/510

I'm just submitting code changes.

Please check line no 593 for currentUrl. 

Thanks to Q2A user Dan D, and Github users valan, iamdand, cksagherian for this patch.

### DIFF
--- a/qa-plugin/facebook-login/base_facebook.php
+++ b/qa-plugin/facebook-login/base_facebook.php
@@ -372,18 +372,17 @@ abstract class BaseFacebook
       return false;
     }
 
-    $response_params = array();
-    parse_str($access_token_response, $response_params);
-
-    if (!isset($response_params['access_token'])) {
+    $response = json_decode($access_token_response);
+    if (!isset($response->access_token)) {
       return false;
     }
 
     $this->destroySession();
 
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for this! Worked fine for me (once I installed the curl extension...) I'll try and get Gideon to update the Q2A site asap.

I'll also take a closer look at the plugin as a whole in the future. I'm not really a fan of the JavaScript widget that Facebook provides, I've regularly had issues with it (for example making an infinite loop). I think it will be better to have something similar to Stack Overflow and other sites where we have a plain button and do the login server-side instead of with popups.

ProThoughts:
> I think it will be better to have something similar to Stack Overflow and other sites where we have a plain button and do the login server-side instead of with popups.
It is good idea. Also please add support for Google Plus login. There are many users who uses gmail for sign up. thanks!

PR #513, opened 2017-04-04
Updated PHPMailer to v5.2.23 (from v5.2.14)

ProThoughts:
Updated PHPMailer to v5.2.23  (from v5.2.14) stable version. There are some security fixes in this version. Please see https://github.com/PHPMailer/PHPMailer

These two files are taken from above repository. 

**Security notices relating to PHPMailer**
https://github.com/PHPMailer/PHPMailer/blob/master/SECURITY.md
Taken from above file....
PHPMailer versions prior to 5.2.22 (released January 9th 2017) have a local file disclosure vulnerability, CVE-2017-5223. If content passed into msgHTML() is sourced from unfiltered user input, relative paths can map to absolute local file paths and added as attachments. Also note that addAttachment (just like file_get_contents, passthru, unlink, etc) should not be passed user-sourced params either! Reported by Yongxiang Li of Asiasecurity.

PHPMailer versions prior to 5.2.20 (released December 28th 2016) are vulnerable to CVE-2016-10045 a remote code execution vulnerability, responsibly reported by Dawid Golunski, and patched by Paul Buonopane (@Zenexer).

PHPMailer versions prior to 5.2.18 (released December 2016) are vulnerable to CVE-2016-10033 a remote code execution vulnerability, responsibly reported by Dawid Golunski.

### DIFF
--- a/qa-include/vendor/PHPMailer/class.phpmailer.php
+++ b/qa-include/vendor/PHPMailer/class.phpmailer.php
@@ -31,7 +31,7 @@ class PHPMailer
      * The PHPMailer Version number.
      * @var string
      */
-    public $Version = '5.2.14';
+    public $Version = '5.2.23';
 
     /**
      * Email priority.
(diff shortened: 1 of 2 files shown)


svivian:
Thanks. I don't usually update third-party libraries until a new Q2A version is about to be released (because it could update again in the meantime) but this one is worthwhile.

ProThoughts:
I'm setting up test site so I feel we should update third-party libs regularly to latest stable version. It will help to detect issues early (if any).

svivian:
Possibly yes, but we could also spend time trying to fix or work around something that gets fixed long before v1.8 is released.

PR #521, opened 2017-04-26
Resolve "A non-numeric value encountered" error on file format.php

christianrss:
The error "A non-numeric value encountered" ocurrs when has no questions in database.
Solved with a explicit type convertion.


svivian:
Thanks, would you be able to redo this against the 1.8 branch? There have been some changes that will conflict with any changes to the dev branch.

christianrss:
yes, i have another fixs to send, so i will create then another pull request.

PR #522, opened 2017-04-26
Fixed problems with a types and refactored the code of files sizes conversions

christianrss:
Fixed some problems like "A non-numeric value encountered" which occurs with a values without explicit type conversion. And refactored the code of file sizes conversions with a better approach. In this branch the field "Maximum size of uploads" presented problems with float numbers what has been fixed with the refactoring.

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -1206,7 +1206,7 @@ function qa_html_page_links($request, $start, $pagesize, $count, $prevnext, $par
 	if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
 
 	$thispage = 1 + floor($start / $pagesize);
-	$lastpage = ceil(min($count, 1 + QA_MAX_LIMIT_START) / $pagesize);
+	$lastpage = ceil(min((int)$count, 1 + QA_MAX_LIMIT_START) / $pagesize);
 
 	if ($thispage > 1 || $lastpage > $thispage) {
 		$links = array('label' => qa_lang_html('main/page_label'), 'items' => array());
--- a/qa-include/qa-base.php
+++ b/qa-include/qa-base.php
@@ -1284,7 +1284,16 @@ function qa_post_limit_exceeded()
 */
 function convert_to_bytes($unit, $value)
 {
-	return intval($value) * 1024 ** (int) array_search(strtolower($unit), ['', 'k', 'm', 'g']);
+	switch (strtolower($unit)) {
+	    case 'g':		
+		return $value * pow(1024, 3);		
+	    case 'm':		
+		return $value * pow(1024, 2);		
+	    case 'k':		
+		return $value * 1024;		
+	    default:		
+		return $value;
+	}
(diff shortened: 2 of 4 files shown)


qa-include/qa-base.php
```diff
@@ -1284,16 +1284,7 @@ function qa_post_limit_exceeded()
...
-		case 'g':
-			return $value * 1073741824;
-		case 'm':
-			return $value * 1048576;
-		case 'k':
-			return $value * 1024;
-		default:
-			return $value;
-	}
+	return intval($value) * 1024 ** (int) array_search(strtolower($unit), ['', 'k', 'm', 'g']);
```
svivian: This feels like it's trying to be too clever, and it makes the code very non-obvious. I know the current code uses somewhat "magic" numbers but it's simple and clear.

qa-plugin/wysiwyg-editor/qa-wysiwyg-editor.php
```diff
@@ -113,9 +113,9 @@ public function admin_form(&$qa_content)
 				array(
 					'id' => 'wysiwyg_editor_upload_max_size_display',
 					'label' => 'Maximum size of uploads:',
-					'suffix' => 'MB (max '.qa_html(qa_format_number($this->bytes_to_mega(qa_get_max_upload_size()))).')',
+					'suffix' => 'MB (max '.qa_html(number_format($this->bytes_to_mega(qa_get_max_upload_size()), 1)).')',
 					'type' => 'number',
-					'value' => qa_html(number_format($this->bytes_to_mega(qa_opt('wysiwyg_editor_upload_max_size')))),
+					'value' => qa_html(number_format($this->bytes_to_mega(qa_opt('wysiwyg_editor_upload_max_size')), 1)),
```
svivian: Is there a reason to have decimals here?

svivian:
This also doesn't appear to have the change from your previous PR.

qa-plugin/wysiwyg-editor/qa-wysiwyg-editor.php
christianrss: Before this change it was not possible put float numbers in maximum file size, for example 1.5 MB.

qa-include/app/format.php
```diff
@@ -1206,7 +1206,7 @@ function qa_html_page_links($request, $start, $pagesize, $count, $prevnext, $par
 	if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
 
 	$thispage = 1 + floor($start / $pagesize);
-	$lastpage = ceil(min($count, 1 + QA_MAX_LIMIT_START) / $pagesize);
+	$lastpage = ceil(min((int)$count, 1 + QA_MAX_LIMIT_START) / $pagesize);
```
christianrss: Here is the correction sent in the previous PR

svivian:
Sorry for the long delay, finally got around to merging this (needed to fix some whitespace issues so I squashed to one commit). Thanks for the work!

PR #523, opened 2017-04-29
Updated to CKEditor 4.6.2

ProThoughts:
Updated to latest CKEditor v4.6.2. This plugin is tested on http://demo18.question2answer.info, it is live there now.

Procedure followed to update as mentioned in below file.
https://github.com/q2a/question2answer/blob/1.8/qa-plugin/wysiwyg-editor/CUSTOMIZE.md

It is better to keep all third party components updated regularly(for security and latest features), someone may find issue while testing q2a dev version.


gurjyot:
Finally Ckeditor is working on mobile :)

But, can't we have an option where users can choose between basic, standard and full features of Ckeditor? It would be amazing to have that feature.

ProThoughts:
@gurjyot, will check that. 

I'm thinking to add math, code editor etc default options in plugin and release separate plugins. It may help someone.

svivian:
Thanks for taking the effort, but as I said before it's not worth updating the dependencies yet as they will update again before too long. Will look at this again when we get closer to a release. Feel free to keep it updated on your demo site, that will be useful for checking any bugs.

@gurjyot There are so many options for CKEditor I think it's best to allow site owners to customize their own builds as described in `CUSTOMIZE.md`

PR #527, opened 2017-06-16
Fixed php 5.3 Travis CI fail issue

ProThoughts:
Travis CI failing for Q2A v18, issue with php 5.3, see below link for more detail
http://question2answer.org/qa/58152/travis-ci-failing-for-q2a-v18-issue-with-php-5-3


svivian:
Thanks for this!

Issue #536, opened 2017-08-13
[Bug] Comment upvote is using question string


ProThoughts:
More details here, issue reported by Ronlaw
http://meta.question2answer.info/569/translation-comment-upvote-is-using-question-string

pupi1985:
It might be better just to merge the QAC message similarly to the `vote_disabled_hidden_x` messages. That way you have less language strings to care about.

ProThoughts:
@pupi1985, agree with you.

ProThoughts:
Fixed this issue in my branch, testing now.

ProThoughts:
Oh, just seen your commit, https://github.com/q2a/question2answer/pull/537 , I will not add mine. thanks for the quick fix.

PR #537, opened 2017-08-13
Unify own post message language string

pupi1985:
Just passing by and saw the simple issue #536 so I thought of sending a keep-alive.

I suggest deprecating `vote_disabled_my_x the same` the same as `vote_disabled_hidden_x`.


svivian:
Good stuff, thanks.

PR #539, opened 2017-08-26
Fix a few warnings generated by reading contents from non-existent files

pupi1985:
At least happening in PHP 7. Applied the old-style fix because it is deprecated code.


svivian:
Makes sense, thanks.

PR #543, opened 2017-09-30
Fix table counts during Q2A installation

pupi1985:
To reproduce this create a new schema. Add a custom table. Start installing Q2A. You'll eventually see:
image

The fix just filters based on the Q2A table prefix. It also takes that into account when upgrading as it doesn't make any sense to block non-Q2A tables while doing so.

Note plugins should still receive the full table list (not the filtered one). Otherwise, this won't be backwards compatible.

This makes me wonder whether plugins should use or not the Q2A table prefix in their table names. I used to think they shouldn't in order to avoid "namespaces" issues with the core. However, other issues might still arise if there are 2 sites sharing the same schema using different prefix that have the same plugin installed. That plugin would try to create the same table twice.

I'm even thinking plugins should have specific sub-prefix for table names so that the table names look like this `qa_plugin_*`. So the core will know it can't use that prefix for a table name nor take them into account when counting them. Not for 1.8 I guess :)

### DIFF
--- a/qa-include/db/install.php
+++ b/qa-include/db/install.php
@@ -555,7 +555,7 @@ function qa_array_to_keys($array)
  */
 function qa_db_missing_tables($definitions)
 {
-	$keydbtables = qa_array_to_keys(qa_db_list_tables());
+	$keydbtables = qa_array_to_keys(qa_db_list_tables(true));
 
 	$missing = array();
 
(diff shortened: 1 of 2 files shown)


svivian:
I'm not able to reproduce this in 1.7.5 or 1.8. What are the specific circumstances? By "create a new schema" I presume you mean a new database? I did that, added a table manually then ran the install and it worked fine.

svivian:
OK. I think this is a good PR anyway - it makes sense to provide only the tables for the current site.

pupi1985:
OK. Just reopened it.
I was wondering, should PR's go now to `dev` now or should we wait for the official 1.8 release?
Also, do you want me to rebase this into `dev` rather than `1.8`?

svivian:
Yes I pushed some commits to the dev branch so it's probably easier to use that for now. No worries about this PR, I merged it into dev. Thanks!

PR #547, opened 2017-10-06
Updated Travis build status image link

ProThoughts:


svivian:
Thanks. I guess we can't point to different Travis images on different branches because they'll keep overwriting each other.

PR #551, opened 2017-10-12
Fix non-ANSI group-by queries

pupi1985:
There were a few missing ones. I haven't tested this much but the fix should be quite mechanical: if the ordering didn't originally care, then getting any of them should be the same.

### DIFF
--- a/qa-include/db/hotness.php
+++ b/qa-include/db/hotness.php
@@ -61,7 +61,7 @@ function qa_db_hotness_update($firstpostid, $lastpostid = null, $viewincrement =
 	if (!isset($lastpostid))
 		$lastpostid = $firstpostid;
 
-	$query = "UPDATE ^posts AS x, (SELECT parents.postid, parents.created AS qcreated, COALESCE(MAX(children.created), parents.created) as acreated, COUNT(children.postid) AS acount, parents.netvotes, parents.views FROM ^posts AS parents LEFT JOIN ^posts AS children ON parents.postid=children.parentid AND children.type='A' WHERE parents.postid BETWEEN # AND # GROUP BY postid) AS a SET x.hotness=(" .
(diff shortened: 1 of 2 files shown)


svivian:
Where exactly are these causing problems? I fixed the hotness query in https://github.com/q2a/question2answer/commit/f547f4289f3e2babcd5c519fad10e14852e34ac7 and I haven't had any more problems.

pupi1985:
This came from this post http://www.question2answer.org/qa/60209
Whenever that piece of code is executed I faced that error too (after setting the appropriate SQL mode flag). Accessing a question generated the error immediately (same as Arjun mentioned) refreshing didn't generate it back. I guess that's because the hotness got already updated by my IP address and didn't fire again.

I don't remember if I faced the one in `selects.php` but it would require the function to be called with the `$full` variable set to `true`.

Regarding the fix you did to the `hotness.php` it seems to be incomplete. Every field in a SELECT statement must be either an aggregate function or part of the GROUP BY. You turned one into an aggregate function but not the rest. I think shouldn't be working in your end either. Can you confirm that?

A side note that I'm actually realizing after seeing the `COUNT` aggregate function is that, although it fixes the syntactical error it might be semantically wrong. For example:

Non-grouped query: `SELECT postid, acount FROM table`

| postid  | acount |
| --- | --- |
| 1 | 15 |
| 1 | 27 |
| 2 | 6 |

Grouping by in the non-ANSI way: `SELECT postid, acount FROM table GROUP BY postid`

Note MySQL is just keeping one of the `acount` values and it is unpredictable which one is going to keep. I chose 15 in this case:

| postid  | acount |
| --- | --- |
| 1 | 15 |
| 2 | 6 |

Grouping by in the an ANSI way using MIN/MAX: `SELECT postid, MAX(acount) FROM table GROUP BY postid`

Note MySQL is being forced to keep the highest `acount`. We're still taking one of them the only difference is that at least we can predict which one. It could also be a `MIN()`.

| postid  | acount |
| --- | --- |
| 1 | 27 |
| 2 | 6 |

Grouping by in the an ANSI way using COUNT: `SELECT postid, COUNT(acount) FROM table GROUP BY postid`

Note in this case we are not keeping any of the original records but rather we're counting them. So this is changing the logic from taking an unpredictable value to counting the values.

| postid  | acount |
| --- | --- |
| 1 | 2 |
| 2 | 1 |

So I think the `COUNT` should be changed to a `MIN` or a `MAX`.

svivian:
Still haven't been able to reproduce the problem but it fixed the problem for Arjun so should all be good :)

PR #562, opened 2017-11-24
Paste/Paste isn't working in ckEditor

gurjyot:
These buttons are just showing and not doing anything.

https://github.com/ckeditor/ckeditor-dev/issues/595


svivian:
Does it work in any browser? Not working in Chrome (desktop) for me, but that CKE issue is talking about mobile. The problem with removing the paste buttons is that CKE gives a message when the paste button is clicked, saying to use Ctrl+V. So if a user cuts some text using the button and there is no paste button they might not know to use Ctrl+V like the message says.

Maybe it's worth removing copy/paste entirely?

gurjyot:
Yes it's worth it to remove it entirely for now. Ckeditor might bring back this feature back in future for mobile as there is a backlash (since mobile users are a big deal nowadays). Till then it's ok to remove it, rather than confusing people.

PR #563, opened 2017-12-05
Change default theme setting to SnowFlat

ProThoughts:


svivian:
Thanks for the PRs (and sorry for the delay in getting to them). Makes sense to have the SnowFlat theme as default now. I held off because I wasn't 100% happy with a few parts like the users page but I never got around to fixing it.

ProThoughts:
No problem. I agree with you. 

Its a time we should think about coming up with completely new theme for q2a. design a better user page also, like stackoverflow.

PR #570, opened 2017-12-17
Added anchor to be able to link to admin plugins

q2apro:
Adds a link icon in front of each plugin. Very helpful if you need to quickly jump to the plugins from other pages. The qa_admin_plugin_options_path() changes with each new installation. This implementation gives a unique and unchanging link. Feel free to change the icon.

Examples:
`/admin/plugins#tagging-tools`
`/admin/plugins#prevent-simultaneous-edits`

image

### DIFF
--- a/qa-include/pages/admin/admin-plugins.php
+++ b/qa-include/pages/admin/admin-plugins.php
@@ -194,7 +194,9 @@ if (!empty($fileSystemPlugins)) {
 		$beforeDbInit = isset($metadata['load_order']) && $metadata['load_order'] === 'before_db_init';
 		$enabled = $beforeDbInit || !$allowDisable || in_array($pluginDirectory, $enabledPlugins);
 
-		$pluginhtml = $namehtml . ' ' . $authorhtml . ' ' . $updatehtml . '<br>';
+		$slugid = qa_slugify($metadata['name']);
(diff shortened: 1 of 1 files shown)


svivian:
> The qa_admin_plugin_options_path() changes with each new installation.

Are you talking about when you have multiple different websites? So you're on `example.com/site/admin#plugin` and you want to go to `example.com/othersite/admin#plugin` ?

q2apro:
In my case I have an admin dropdown where I can quickly access the plugins and their options, e. g. Featured Questions, Stop Spam etc.

svivian:
In that case, can't you get the required hashes for your dropdown from `qa_admin_plugin_directory_hash` or `qa_admin_plugin_options_path` ?

q2apro:
As I said, the hash changes with each new installation. So when I have a bookmark in my browser, click this, then it is not showing the plugin. It would be great if the hash / show=... could be constant.

svivian:
The hash should be constant, as long as the plugin directory stays the same. However I think it may have changed once or twice, ironically due to fixing bugs with how they are generated or linked. They are now "normalized" so they should not change in the future.

q2apro:
If the do not change in the future, then it would be great and the fix is obsolete!

PR #581, opened 2017-12-25
A few controller's branch comments

pupi1985:
Really not much to comment on this. I didn't quite take a look at the whole branch but rather just to the Router class. The changes I made are relatively minor.

I have some comments:
 1. Is there any advantage in using `'\Q2A\Controllers\User\UserProfile'` rather than `UserProfile::class`? I've seen many people using that and I just don't see any advantage rather than loosing IDE support (e.g. autocompletion or a rename refactor)
 1. Related to the previous question, is there any reason not to `use` namespaced classes?
 1. I think this branch has a stronger need of a global "framework" object. I mean, an object that holds the different components, e.g., the router. It could be possible to have a singleton router, however, static methods usually bring issues when trying to apply dependency injection later so I'd rather use one main singleton object that references the other (detachable) subcomponents. There as a related PR I made a while ago, most likely useless today: https://github.com/q2a/question2answer/pull/211/commits

### DIFF
--- a/qa-include/app/options.php
+++ b/qa-include/app/options.php
@@ -645,7 +645,7 @@ function qa_message_html_defaults()
 
 /**
  * Return $voteview parameter to pass to qa_post_html_fields() in /qa-include/app/format.php.
- * @param $postorbasetype The post, or for compatibility just a basetype, i.e. 'Q', 'A' or 'C'
+ * @param string|array $postorbasetype The post, or for compatibility just a basetype, i.e. 'Q', 'A' or 'C'
  * @param bool $full Whether full post is shown
  * @param bool $enabledif Whether to do checks for voting buttons (i.e. will always disable voting if false)
  * @return bool|string Possible values:
--- a/qa-include/app/page.php
+++ b/qa-include/app/page.php
@@ -193,7 +193,7 @@ function qa_get_request_content()
 		qa_set_template($route->getId());
 		$controllerClass = $route->getController();
 		$ctrl = new $controllerClass();
-		$qa_content = call_user_func_array(array($ctrl, $route->getFunction()), $route->getParameters());
+		$qa_content = $ctrl->executeAction($route->getAction(), $route->getParameters());
 
 	} elseif (isset($routing[$requestlower])) {
 		qa_set_template($firstlower);
--- a/qa-include/app/page.php
+++ b/qa-include/app/page.php
@@ -19,6 +19,13 @@
 	More about this license: http://www.question2answer.org/license.php
 */
 
+use Q2A\Controllers\User\UserMessages;
+use Q2A\Controllers\User\UserPosts;
+use Q2A\Controllers\User\UserProfile;
+use Q2A\Controllers\User\UsersList;
+use Q2A\Http\Route;
+use Q2A\Http\Router;
+
 if (!defined('QA_VERSION')) { // don't allow this page to be requested directly from browser
 	header('Location: ../../');
 	exit;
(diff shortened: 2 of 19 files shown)


svivian:
Thanks for the PR and comments. To answer your questions:

1. The main reason is because `::class` is only available in PHP 5.5+ and Q2A 1.9 will be targeting 5.3 or 5.4 minimum (they still command a significant market share). Even so, if we are using the full namespace anyway, there’s not much reason to use `::class` since we typed the whole string already.
2. There’s no specific reason. I generally feel that using a `use` when you only reference the class once (particularly in a huge file like qa-base.php) is a bit pointless. With the controllers that would massively bloat the `use`s when we add all paths, just to be used in one function. (However if we did use `::class` it could make sense to use a ‘group’ like `use Q2A\Controllers\User`.)
3. Totally agree, that’s why I kept that PR around :) I still have it as a local branch. What you’ve described above sounds more like a ‘Service Locator’ pattern - I’d be happy to resurrect the PR as that if you like. Would make it easier to fetch objects for DI when necessary.

I guess the global variable thing from that PR would be a slightly different concern, which could still be part of the ‘Application’ class. 

Will add a few line notes.

pupi1985:
That's basically it. My intention was to make some comments and suggestions but writing them take 70% of the time that takes me to actually add the code so, in practice, it is better to code them and then make smaller discussions on the unclear or disagreed items.

Regarding the Service Locator pattern, that's right. That is a better way of saying "Globally accessible bag of stuff" :D As long as it doesn't go out of control, it should be fine.

Again, these were just suggestions and ideas. If you'd like to actually merge this I can make any needed change (e.g. removing the `::class` references). Or just sleep on it and see if you think in a better approach.

svivian:
Happy to merge this if you make the few changes above (remove `::class`, use `strtr` function, update the Exception namespaces). This is full of great stuff!

I'll try and finish the Database stuff I was working on in the next few days and we can look at adding a Service Locator and/or Application class to hide the globals.

pupi1985:
I got the impression the DB stuff was a WIP so I didn't quite dug deep into it.

I've made the changes we discussed and added the commits to the branch. I considered rebasing the branch but on one hand that would mean you having to review everything again (rather than just the changes) and, on the other hand, I don't really know what would happen to the comments on the commits that would no longer exist... according to Back to the Future, they should fade out... but I rather not find out :)

Oh, and before I forget Happy New Year! :tada:

svivian:
Yes the DB is in progress. I've been playing with a DbConnection class (thin wrapper for PDO) and DbResult class (wrapper for PDOStatement). And a move towards using built-in prepared statements instead of the `$`/`#` parameters we have now.

Not sure about some of the incidental functions in `qa-db.php` like `qa_db_random_bigint` or
`qa_db_list_tables` - they don't really fit in DbConnection. Perhaps a DbUtils class? If you have any ideas let me know.

I've merged the PR now, thanks! And Happy New Year to you too 🎆  (apparently that's a fireworks icon...)

pupi1985:
`qa_db_random_bigint` is a big tricky as it doesn't really return an int but rather a string. It could be part of `util/string.php` too.

`qa_db_list_tables` is clearly related to the DB. I get the impression that it should be part of a class related to DDL. For example, there could be a class `Schema` that would simplify some basic behavior for creating or altering the database or tables structure. Something similar to what the install/upgrade processes use. Having this function (method) in there would make sense to me.

I'll be taking a look at the classes you uploaded. If I later find something to suggest, I'll create a new PR

svivian:
`qa_db_list_tables` would fit with some of the functions in `db/admin.php` like `qa_db_table_size`.

I was thinking we'll probably use some kind of Repository pattern for the `db/*.php` functions (e.g. a UserRepo for finding a user in the DB etc). So those admin functions would go in a generic DbRepo or MetaRepo, for functions about the DB itself.

PR #584, opened 2018-01-08
Some database and a few controllers thoughts

pupi1985:
I tried to make some independent commits so that you can cherry pick whatever you like.

**The IN statement**
https://phpdelusions.net/pdo#in
https://medium.com/coding-design/preparing-for-prepared-statements-619292bc4d8a

This makes me feel I'm in back in 1999. How I see it, the main issue is knowing which parameter placeholders are arrays and which ones are not. I don't really like the approach I've implemented but I don't like any other either.

**UserPosts**
I don't really know if what I did to this file makes it more or less readable. Feel free to get rid of the suggested changes.

### DIFF
--- a/qa-src/Controllers/User/UserPosts.php
+++ b/qa-src/Controllers/User/UserPosts.php
@@ -30,6 +30,12 @@ class UserPosts extends \Q2A\Controllers\BaseController
 	protected $userid;
 	protected $userhtml;
 
+	/**
+	 * @param string $handle
+	 *
+	 * @return array
+	 * @throws PageNotFoundException
+	 */
 	public function activity($handle)
 	{
 		$this->userHtml($handle);
@@ -99,6 +105,12 @@ class UserPosts extends \Q2A\Controllers\BaseController
 		return $qa_content;
 	}
 
+	/**
+	 * @param string $handle
+	 *
+	 * @return array
+	 * @throws PageNotFoundException
+	 */
 	public function questions($handle)
 	{
 		$this->userHtml($handle);
@@ -172,6 +184,12 @@ class UserPosts extends \Q2A\Controllers\BaseController
 		return $qa_content;
 	}
 
+	/**
+	 * @param string $handle
+	 *
+	 * @return array
+	 * @throws PageNotFoundException
+	 */
 	public function answers($handle)
 	{
 		$this->userHtml($handle);
@@ -251,10 +269,14 @@ class UserPosts extends \Q2A\Controllers\BaseController
 		return $qa_content;
 	}
 
+	/**
+	 * Return the HTML to display for the handle, and if we're using external users, determine the userid.
+	 *
+	 * @param string $handle
+	 * @throws PageNotFoundException
+	 */
 	private function userHtml($handle)
 	{
-		// Get the HTML to display for the handle, and if we're using external users, determine the userid
-
 		if (QA_FINAL_EXTERNAL_USERS) {
 			$this->userid = qa_handle_to_userid($handle);
 			if (!isset($this->userid)) { // check the user exists
--- a/qa-src/Controllers/User/UserPosts.php
+++ b/qa-src/Controllers/User/UserPosts.php
@@ -27,9 +27,6 @@ require_once QA_INCLUDE_DIR . 'app/format.php';
 
 class UserPosts extends \Q2A\Controllers\BaseController
 {
-	protected $userid;
-	protected $userhtml;
-
 	/**
 	 * @param string $handle
 	 *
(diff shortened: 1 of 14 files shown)


svivian:
The IN queries is an interesting one. I think what you’ve done is fine, if a bit convoluted. Can’t think of any better way to handle it within the Db class at the moment.

The alternative would be for the code outside the Db class to handle it, with some provided methods, eg

    $sql = '... WHERE field IN (' . $db->makeParameters($userids) . ')';
    $params = array_merge($params, $userids);

pupi1985:
The IN queries are a mess no matter how you look at them. The approach you suggest doesn't provide quite a lot of abstraction. I'd rather sacrifice some performance to increase readability.

Another approach I was thinking about was using named parameters

```php
$db->query('... WHERE field IN (:userids)', array('userids' => array(1, 5, 8)));
```
That would ease replacements such as`:userids1`, `:userids2`, etc. but may be a bit redundant to write. I do get the impression that the named approach is safer: I'm concerned about finding a question mark that is not a parameter... but I couldn't think of any way in which that might happen in MySQL... but if the idea is to change the DBMS eventually we might get some issues... I think.

svivian:
I've cherry-picked some of these commits. With the UserPosts one, there's now a stray query outside the `qa_db_select_with_pending()` call which conflicts with the `QA_OPTIMIZE_DISTANT_DB` setting, so I've left that one.

For some reason the `IN (...)` commits cause a conflict now, I guess because I skipped some commits. I'll take a proper look later.

qa-src/Controllers/User/UserPosts.php
```diff
@@ -27,243 +27,329 @@
...
 
 		// Get information on user questions
 
-		$pagesize = qa_opt('page_size_qs');
-		$count = (int)@$userpoints['qposts'];
-		$questions = array_slice($questions, 0, $pagesize);
-		$usershtml = qa_userids_handles_html($questions, false);
-
+		$pageSize = qa_opt('page_size_qs');
+		$count = (int)@$userPoints['qposts'];
```
Mariusz08: You should remove `@` char before variable, and `(int)` MUST have space after statement

`$count = (int) $userPoints`

qa-src/Controllers/User/UserPosts.php
```diff
@@ -27,243 +27,329 @@
...
+	 * @return mixed|string
+	 */
+	private function userHtml($handle, $userId)
+	{
+		if (QA_FINAL_EXTERNAL_USERS) {
+			$usersHtml = qa_get_users_html(array($userId), false, qa_path_to_root(), true);
 
-			$qa_content['q_list']['qs'][] = qa_other_to_q_html_fields($question, $loginuserid, qa_cookie_get(),
-				$usershtml, null, $options);
+			return @$usersHtml[$userId];
```
Mariusz08: You should use `try..catch` statement

qa-src/Controllers/User/UserPosts.php
```diff
@@ -27,243 +27,329 @@
...
+	}
 
+	/**
+	 * @param string $handle
+	 * @param mixed $loggedInUserId
+	 * @param mixed $userId
+	 *
+	 * @return array
+	 */
+	private function fetchAnswersDataFromDb($handle, $loggedInUserId, $userId)
```
Mariusz08: You can use type-hints

qa-src/Controllers/User/UsersList.php
```diff
@@ -18,13 +18,25 @@
...
 require_once QA_INCLUDE_DIR . 'app/format.php';
 
 class UsersList extends \Q2A\Controllers\BaseController
 {
+	public function __construct()
+	{
+		parent::__construct();
+
+		$this->addMiddleware(new InternalUsersOnly(), array('newest', 'special', 'blocked'));
+		$this->addMiddleware(new MinimumUserLevel(QA_USER_LEVEL_MODERATOR), array('blocked'));
```
Mariusz08: `[]`

qa-src/Controllers/User/UserPosts.php
pupi1985: How would you change the function definition to use type-hints?

qa-src/Controllers/User/UserPosts.php
pupi1985: Exactly what exception are you suggesting to catch in that line?

qa-src/Controllers/User/UserPosts.php
Mariusz08: `private function fetchAnswersDataFromDb($handle, int $loggedInUserId, int $userId): array`

qa-src/Controllers/User/UserPosts.php
Mariusz08: Hmm... 
I did not look exactly at this code, but in the code we should not use `@` - if I see such a sign next to a variable / function, I always try to rewrite the code so that the character does not exist.

qa-src/Controllers/User/UserPosts.php
pupi1985: Would you mind pointing me to a php.net page that states that syntax is valid?

qa-src/Controllers/User/UsersList.php
pupi1985: Remember this branch needs to be PHP 5.3 compatible

qa-src/Controllers/User/UserPosts.php
Mariusz08: http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration - for arguments (PHP 7)

http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration - for return type declaration (PHP 7)

If we works at PHP 5 you cannot use this declarations

PR #586, opened 2018-01-13
Add a simple application and container and update them to use the router

pupi1985:


svivian:
Thanks for this. Made a few minor changes:

- Replaced resolve/bind with get/set in the Container for simplicity.
- Renamed the `app` function to `qa_app` to avoid clashes (even though you used `function_exists`, if someone else uses `app()` it will break a bunch of stuff anyway).
- Split that out into `qa_app` which just gets the app and `qa_service` for getting stuff from the Container.
- Replaced `is_null` with `=== null`. Mainly personal preference, but no need to call functions where not necessary :)

PR #591, opened 2018-01-21
Many random tweaks

pupi1985:


svivian:
Just FYI, I found that some of the CSS was causing problems - the `background` property resets all properties to their defaults, so `background: url(...)` removes any previously defined background colors and works like explicitly setting `background: transparent url(...) repeat 0 0;` 

I've fixed the issues in f9f9d3bb566859204d5d6cef4f71e04e9a4a6c99

PR #592, opened 2018-01-21
Send qa_debug output through qa_html function

pupi1985:
I was having a hard time because I was getting `user name ` from an email address `user name <email@server.com>` when sending it to qa_debug. After a while, I found out that qa_debug was not escaping the HTML

### DIFF
--- a/qa-include/qa-base.php
+++ b/qa-include/qa-base.php
@@ -1867,9 +1867,9 @@ function qa_opt($name, $value = null)
  */
 function qa_debug($var)
 {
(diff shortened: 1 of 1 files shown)


svivian:
Had an issue with this where the output of `htmlspecialchars` was blank. Apparently due to invalid UTF-8 characters, I'm guessing it was the binary IP address (I was printing the whole `$qa_content` var to test). Wasn't sure of the impact of changing the `qa_html` function so I went for a simple change to the `qa_debug` function: a81fbf0209304966d2c6eebce68a9e90814efa97. Thanks for the PR though.

PR #594, opened 2018-01-25
Refactored the Recalc code into a set of classes.

Spudley:
Hi.

I've had less spare time for refactoring than I had hoped, but I have managed to do a bit for you.

This is a refactor of the Recalc code (qa-include/app/recalc.php) into a set of classes.

Hope you like it.


svivian:
Cool, thanks! I'll review this when I get a chance.

Spudley:
Hia. How's it going? Did you have a chance to review yet?  😁

svivian:
Hello, apologies for the delay getting to this, been super busy the past month. I've looked over the changes and the general idea is great!

One thing I'm not sure of is the `AbstractFinalStep` and the various `*Complete` classes. They're basically all identical besides the message, some are completely empty. We should be able to use the `$isFinalStep` variable on the regular classes, right? Then have the complete message in that class, e.g. something like this on `Q2A_Recalc_RecountPostsACount`

	public function getMessage()
	{
		if ($this->isFinalStep) {
			return qa_lang('admin/recount_posts_complete');
		} else {
			return $this->progressLang('admin/recount_posts_as_recounted', $this->state->done, $this->state->length);
		}
	}

Spudley:
Hi. Thanks for the feedback. I'm glad you like it.

I'm going to respond to your point and counter-argue it. But please don't read this as me trying to force the issue; this is your software, not mine, so I'll follow your lead. But I want to try to explain my thought process first so you can see why I've done it this way.

The point of the `*Complete` classes and `AbstractFinalStep` is to reduce the decision points to polymorphic classes. Yes you're right that it could have been done with an `if()` block, but I felt that having them in separate classes is "cleaner". It's a fairly common technique with OO code and clean coding practices to have empty or near-empty polymorphic classes like this.

To be fair, these `*Complete` classes were one of the last things I did; I wasn't intending to have them originally, but it became necessary as I worked to reduce the complexity of the code without changing the actual functionality. There is a 'complete' step for everything; it may not do much, but it is a step in its own right in the existing recalc, so it got a class in its own right in my version.

Changing it the way you suggest wouldn't just add an `if()` block to a bunch of classes, it would also mean that those classes would end up being used for two separate stages, which kinda breaks the class-for-a-stage concept I've gone with. We'd have to add code to tell the class which of those stages it was running by manually setting the `final` flag when we create it, and we'd have to have some way of knowing when to set that. We'd have to decide arbitrarily which of the other stages classes to add the 'complete' message to, which doesn't feel right. Also, doing it that way would make it more difficult to modify any of the processes, eg by adding additional stages in the future.

So cutting down on those tiny classes would actually mean adding complexity elsewhere.

So that's my argument. I'm not going to pretend that my code is perfect. I do think I've taken the right approach here, but as I say, if you don't agree I'll be happy to change it.  :-)


One thing I will say is that I wasn't entirely happy with my choice of naming convention for some of the classes. In particular, I thing `AbstractStep` and `AbstractFinalStep` are slightly clunky names. But I couldn't think of anything better, so there we are. I also considered whether I should have put the stages into subfolders; eg a subfolder for a process like `RefillEvents`, with all the stage classes for that process in that folder. I was in two minds about that, and I'm still not sure. I think it would have been an easier decision if I was writing with namespaces; adding a folder as things stand makes the class names uglier, but with namespaces it's less of an issue.

Thanks again for reviewing. Let me know your thoughts; whether you'd like me to make any changes or not.

All the best.  :-)

svivian:
Thanks for the response. I'll start by saying I disagree with the notion that having empty classes is a "clean OO practice". I've always regarded it as a code smell (and a quick bit of research seems to confirm that - it's apparently known as "lazy class").

*However*, in the case of specific recalc functionality here, I agree the 'Complete' classes make sense within the current structure. But I'd also suggest the current structure isn't the best (that's the fault of the original setup, not your code). For example it shouldn't be the responsibility of a recalc step to know what the next step is, or know whether it's the final step. Ideally we'd have a list of all the steps in `Q2A_Recalc_RecountPosts` and that would execute each one and handle completion. But that doesn't really fit in the current structure. 

So in short, what you've done so far is great. Will merge shortly. We can think about ways to change how the recalc is actually processed later on.

Also regarding subfolders, we'll be using namespaces in the next version so we can do exactly what you suggest. Currently in the 'controllers' branch but I'll sort out naming this properly.

Spudley:
Thanks for the reply, and for accepting the change. :smile:

I'll take a look see what else I can do. (hopefully I'll still have time to work on it; I'm going to be a bit more busy in the near future, but I'll see what I can do)

svivian:
Made a couple of minor changes - the class hierarchy was a little messed up as the various 'Complete' classes were inheriting from `ReindexPostsComplete` instead of `AbstractFinalStep`. I also merged `BlobsToDiskComplete` and `BlobsToDBComplete` into one to avoid the pointless empty class.

Thanks again for your contribution!

BTW I've also merged the work from the 'controllers' branch into 'dev' now, so you can use namespaces for any future additions. The `Q2A` namespace maps to the `qa-src` directory and from there it follows the standard PSR-4 naming.

PR #596, opened 2018-01-26
Clarified the description regarding shared users

q2apro:


svivian:
Thanks! It makes more sense now :)

PR #605, opened 2018-02-10
Fix some scenarios where ^userpoints records are not present and generate issues

pupi1985:
Assuming a user is overriding the `QA_MYSQL_USERS_PREFIX` then there are different sites each with its own `^userpoints` table. The thing is when a site creates a user then that site (and only that site) executes this line: https://github.com/q2a/question2answer/blob/cfec3c0155817ffba49b06435a05ebb3ef9859a1/qa-include/app/users-edit.php#L192

As sites are independent I don't think making a site communicate to each other the creation of the user would make any sense.

That means having to cope with the fact that there might be records in the `^users` table that are not in the `^userpoints` table. I don't think this is a big deal.

The current issue can be seen when listing top users. The query uses an `INNER JOIN` against the `userpoints` table which immediately removes all non present users. The suggestion would be to make a `LEFT JOIN`. It will make queries a bit slower but I don't think there any other way. To compensate, the fix also checks if there are multiple sites sharing the same `^users` table so that the most performant query is used.

### DIFF
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -456,6 +456,17 @@ if (QA_FINAL_EXTERNAL_USERS) {
 				require_once QA_INCLUDE_DIR . 'db/selects.php';
 				$qa_cached_logged_in_user = qa_db_get_pending_result('loggedinuser', qa_db_user_account_selectspec($userid, true));
 
+				// If the site is configured to share the ^users table then there might not be a record in the
+				// ^userpoints table so this creates it
+				if ($qa_cached_logged_in_user['points'] === null) {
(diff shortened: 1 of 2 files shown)


svivian:
Will this also update the user count? I found on my site if people register on different Q2A instances the user counts on each site get out of sync.

pupi1985:
Well, the user count is being updated when the `qa_db_points_update_ifuser()` actually inserts a user record:
https://github.com/q2a/question2answer/blob/cfec3c0155817ffba49b06435a05ebb3ef9859a1/qa-include/db/points.php#L211-L213

Is that what you're talking about?

PR #607, opened 2018-02-16
Simplify hidden HTML and avoid its detection as malware

pupi1985:
Seems to be working based on this comment: http://www.question2answer.org/qa/62349/malware-detected-in-q2a-code?show=62404#c62404

### DIFF
--- a/qa-include/qa-theme-base.php
+++ b/qa-include/qa-theme-base.php
@@ -396,8 +396,7 @@ class qa_html_theme_base
 
 	public function body_hidden()
 	{
-		$indent = $this->isRTL ? '9999px' : '-9999px';
-		$this->output('<div style="position:absolute; left:' . $indent . '; top:-9999px;">');
+		$this->output('<div style="position:absolute;overflow:hidden;clip:rect(0 0 0 0);height:0;width:0;margin:0;padding:0;border:0;">');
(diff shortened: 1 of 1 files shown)


ProThoughts:
cool! thanks for the quick fix.

svivian:
We use absolute positioning so that the image is pre-loaded and appears instantly when doing an action (to show something is happening). Unfortunately browsers don't load images inside `display:none`, so on a slow connection it can take a second for the icon to display, which defeats the purpose.

`visibility:hidden` does load the image but it takes up space on the page. We can add `position:absolute; top:0` but we may be back to square one with that.

I don't think pre-loading via JS is feasible either as the image URL is part of the CSS (different themes have different loading icons).

pupi1985:
I can confirm the browser caching issue. I amended the commit. The solution is overkill on purpose: to avoid any theme CSS change to incidentally make the hidden element visible.

svivian:
OK thanks, hopefully this stops the bogus malware alert.

ProThoughts:
This is working, no malware alert now. I added this patch on below site.
https://sitecheck.sucuri.net/results/meta.question2answer.info

ProThoughts:
@svivian , I feel you should release q2a v1.8.1 because of this patch. thanks!

Issue #574, opened 2017-12-18
Why does Voting response take so much time? Bug?


q2apro:
I realized, and also most of my users, that the Ajax-Voting takes a long time (2 to 5 seconds). All my system is running fast, so I doubt that this is a server issue. 

![image](https://user-images.githubusercontent.com/6764548/34113496-90af0d04-e418-11e7-8d28-a5a813231483.png)

Could it be that we have a code problem here? I remember that we return the full HTML element. I rather would return a number (votes+1) or (votes-1) when downvote. And then change the number frontend. Much easier - and for sure faster.

phy25:
Returning the full HTML alone shouldn't take this long time. Rather we should look into the breakdown of the PHP response time.

svivian:
HTML is returned because every theme could have completely different HTML for the voting (and that HTML could be different for every user if you wanted). So it's not just a case of some JavaScript adding a class.

I've done lots of profiling of Q2A pages but not AJAX requests, so I'll take a look at those and see if there's anything obvious to speed them up.

svivian:
Commenting here for reference. I did some quick testing and found the queries below are quite slow when voting. (You can check this yourself by turning on debug mode and adding `global $qa_usage; $qa_usage->output();` just before the return in `qa-include/ajax/vote.php` and it will show the debug output by the voting buttons.)

	INSERT INTO qa_uservotes (postid, userid, vote, flag, votecreated) VALUES (293661, 1, 1, 0, NOW()) ON DUPLICATE KEY UPDATE vote=1, voteupdated=NOW()

	25.92 ms

	UPDATE qa_posts AS x, (SELECT COUNT(*) AS acount, COALESCE(GREATEST(MAX(netvotes), 0), 0) AS amaxvote FROM qa_posts WHERE parentid=293566 AND type='A') AS a SET x.acount=a.acount, x.amaxvote=a.amaxvote WHERE x.postid=293566

	23.14 ms

	INSERT INTO qa_options (title, content) SELECT 'cache_unupaqcount', COUNT(*) FROM qa_posts WHERE type = 'Q' AND amaxvote = 0 AND closedbyid IS NULL ON DUPLICATE KEY UPDATE content = VALUES(content)

	219.86 ms

[...]

PR #608, opened 2018-02-19
Use JSON in admin_click AJAX calls

pupi1985:
Just me again doing some controversial commits.

The idea was to keep it backwards compatible. I believe I'm just missing the language strings in that sense but I'd like to hear thoughts whether this whole change and the approach makes sense or not before moving on with this.

PS 1: there is a `cache_hiddencount` setting missing that will have to be implemented some day.
PS 2: once all AJAX requests are turned into JSON there is no need to track the `version` of the request as it doesn't actually provide any backwards compatibility.


svivian:
Sweet, this was something that's been on my list for a while now! What is the general format of the response? I see a lot of keys in the `qa_admin_single_click_array` function.

The main AJAX requests from the front end (upvoting, hiding, etc) currently return HTML so presumably there will be a general `html` key for that?

pupi1985:
> What is the general format of the response?

response - success:
```
{
    result: "success"
    <any additional field relevant for each specific response>
}
```
response error:
```
{
    result: "error",
    error: {
        type: string,
        message: string
    }
    <any additional field relevant for each specific response>
}
```
In these particular calls, I'm returning something like this:
```
{
    result: "error",
    error: {
        type: string,
        message: string
    },
    entityCount: integer, // used to update the count
    hideEntitySelector: string // used to hide the clicked element (or not)
}
```
I'm always returning a 200 status code... I said "JSON", not "REST"... if you want the latter, I'll have to charge you separately :)

Note the `error.type` doesn't really have much use but I'm pretty sure someone will build a theme and will want to check the type of the error returned and will go crazy. Even the `message` would be of no use in this case as it can be internationalized.

> The main AJAX requests from the front end (upvoting, hiding, etc) currently return HTML so presumably there will be a general html key for that?

I intentionally avoided any theme output as I didn't want to fall in that discussion [again](https://github.com/q2a/question2answer/issues/574#issuecomment-356499415)... but yes, a decision will have to be made. I'm more inclined to my linked suggestion there but of course it will be more backwards compatible returning the HTML in a key of the JSON response. Probably, for a first step on this it may make more sense to to go for the latter, right?

So do you want me to turn into JSON some other calls or do you prefer to wait to review this? If you want me to go on with this, should I create different PRs or should I keep pushing to the 104 branch?

svivian:
Yes I think it will be best to stick to returning HTML in the response. I think are too many permutations possible in themes to have the hidden HTML templates. 

Is the ‘hideEntitySelector’ key used in your PR? It’s used in updating the `entityCount` but not for the use you describe of hiding things.

On that topic, I think we could use a more generic structure for HTML updates. For example return a map of class selectors and the values/HTML to put in them. Something like:

    dom_updates: [
        {
            selector: ".qa-nav-sub-counter-approve",
            html: "42",
        },
        {
            selector: "#a123 .qa-voting",
            html: "<div class=\"qa-voting-etc\">...</div>"
        }
    ],

What do you think?

pupi1985:
> Is the ‘hideEntitySelector’ key used in your PR? It’s used in updating the entityCount but not for the use you describe of hiding things.

It is because the `qa_conceal` call is wrapped in the `if`: 
https://github.com/q2a/question2answer/pull/608/files#diff-2aee63e040387c69b3e83098633f3913R124

I also like the more generic approach for updating HTML. We might be needing something else too. Something to define the kind of update. E.G.: is it a DOM replacement or an append? I don't want to add too much stuff from the beginning but probably that original JSON structure for the `dom_updates` might evolve a bit.

pupi1985:
Here is the next iteration. There are some common additional actions to bear in mind such as concealing or revealing elements (instead of just replacing them).
I also added the `response.error.severity` property to define if the error is severe enough to show the alert or not.

qa-content/qa-global.js
```diff
@@ -152,22 +152,31 @@ function qa_favorite_click(elem)
...
-		var headerpos = response.indexOf(header);
+	$.extend(params, {qa: 'ajax', qa_operation: operation, qa_root: qa_root, qa_request: qa_request, version: version});
 
-		if (headerpos >= 0)
-			callback(response.substr(headerpos + header.length).replace(/^\s+/, '').split("\n"));
-		else
-			callback([]);
+	$.post(qa_root, params, function (response) {
+			if (version === 0) {
+				var header = 'QA_AJAX_RESPONSE';
```
Mariusz08: You should use `let` or `const` to define javascript variable

pupi1985: As of now, the core does not use ES6. So `should` is arguable :)

svivian:
Just want to let you know I haven't forgotten about this! Haven't had time to dive into it yet, sorry.

pupi1985:
We all have a life, don't we? If you think about it, there is significant activity taking into account this is an unpaid night job :)

qa-content/qa-admin.js
```diff
@@ -112,6 +112,27 @@ function qa_mailing_start(noteid, pauseid)
 	);
 }
 
+function qa_update_dom(response)
+{
+	if (!response.hasOwnProperty('domUpdates')) {
+		return;
+	}
+
+	for (var i = 0; i < response.domUpdates.length; i++) {
```
Mariusz08: Please use `let`

qa-include/app/admin.php
```diff
@@ -581,6 +581,208 @@ function qa_admin_single_click($entityid, $action)
...
+								'selector' => '#p' . $entityid,
+								'action' => 'conceal',
+							),
+						);
+					}
+					break;
+
+				case 'hide':
+				case 'clearflags':
+					$entityCount = (int)qa_opt('cache_flaggedcount');
```
Mariusz08: `(int) qa_opt...` after int MUST be space

qa-content/qa-admin.js
pupi1985: I have already answer you about `let` in another comment which you didn't reply

Mariusz08: Sorry, I don't saw it :) But, you can think to rewrite code to ES6;)

PR #615, opened 2018-03-05
updated travis-ci build image in bugfix branch

ProThoughts:


svivian:
Thanks for the suggestion, but unfortunately we can't have different buttons on different branches because they will cause merge conflicts (e.g. when bugfix gets merged into master for Q2A 1.8.1).

PR #619, opened 2018-03-10
Fix anonymous user images not being displayed

pupi1985:
The function is returning too early so anonymous users don't get to be assigned a default image in question lists. The default image should be displayed regardless of the user being anonymous or not. The link to the user profile should not be displayed for anonymous users.

### DIFF
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -650,9 +650,6 @@ if (QA_FINAL_EXTERNAL_USERS) {
 		if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
 
 		require_once QA_INCLUDE_DIR . 'app/format.php';
-		if (strlen($handle) == 0) {
-			return null;
-		}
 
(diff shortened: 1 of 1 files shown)


qa-include/app/users.php
```diff
@@ -665,6 +662,9 @@ function qa_get_user_avatar_html($flags, $email, $handle, $blobId, $width, $heig
 				break;
 			case 'local-default':
 				$html = qa_get_avatar_blob_html(qa_opt('avatar_default_blobid'), qa_opt('avatar_default_width'), qa_opt('avatar_default_height'), $size, $padding);
+				if (strlen($handle) == 0) {
```
Mariusz08: You can use `===`

pupi1985: Why would I check for type when `strlen` always return an `int`?

Mariusz08: If `strlen` always return `int`, you can use it without stress ;) And seriously, it's better to always use this operator wherever we can

pupi1985: My rule of thumb is: for operands that CAN be of more than one type (mixed) or null always use `===`. Otherwise, `==`, to avoid the unnecessary type-check

PR #620, opened 2018-03-12
Added placeholder "Search..." for search box

ProThoughts:
On mobile search box looks confusing if it is for search or ask a question so added this placeholder.
screen shot 2018-03-11 at 8 45 30 pm

After adding placeholder it looks like this.
screen shot 2018-03-11 at 8 48 43 pm


svivian:
Thanks for this, however we'll need to use the language file for the "Search" text. I think there should be an existing string we can use.

ProThoughts:
agree with you. submitting another patch.

PR #625, opened 2018-03-21
Fixed some error 500 issues on the new recalc approach

pupi1985:
I also upgraded the code to support namespaces.

BTW, shouldn't the hierarchy remove the switch?

https://github.com/q2a/question2answer/blob/c3e5f3ae06084dbf2198342da5482d68e5ee145e/qa-include/Q2A/Recalc/State.php#L133-L140

I would have expected each step to know how to calculate their own length.


Spudley:
Hi. I wrote these changes to recalc, so I'll answer the questions you raised.

I agree that the switch would ideally be removed, but my intention was for this to be a first pass at refactoring it; I kept the functionality, the processes, and the code as close to the original as possible while moving to a class hierarchy, specifically to avoid breaking anything and to make the transition to the new code easy. There are plenty of other ways it could be improved as well.

As I say, it was a first pass. It isn't perfect. It can - and should - be reworked again to improve it further.

pupi1985:
Hi @Spudley
> It can - and should - be reworked again to improve it further.

I'd say that rework is exactly what I did in this PR. Furthermore, the question about the switch just opens the door for the next rework.

PR #628, opened 2018-04-04
Refactor emails.php to classes

Spudley:
Hi.

I have refactored the email.php file into classes. I hope you like what I've done.

Notes:

- I've stubbed out the existing functions in emails.php so that all they do is call the new code, but I've left them there as removing them would break any plugins that do function overrides on them.

- The new code is largely the same as the existing code, with some tidying up done. Obviously re-arranged a bit to suit the new class structure, but other than that I didn't change much.

- I can't see that the notification suspend functionality is actually used anywhere in the system, but I have nevertheless ported it across.

- I've replaced calls to qa_email_validate() with  PHPMailer::validateAddress() as this is a more robust method. It's arguable whether it's even necessary to check this before trying to send as it's one of the first things that PHPMailer does internally anyway.


svivian:
Thanks for this! I will review when I can.

qa-include/Q2A/Notifications/Email.php
```diff
@@ -0,0 +1,139 @@
...
+	 * @param type $handle
+	 * @return \self
+	 * @throws FatalErrorException
+	 */
+	public static function CreateByUserID($userid, $email = '', $handle = '')
+	{
+		if (!(int)$userid) {
+			throw new FatalErrorException('User ID not specified in Notifications/Email/CreateByUserID.');
+		}
+		$needemail = !PHPMailer::validateAddress($email); // take from user if invalid
```
Mariusz08: PSR-2

qa-include/Q2A/Notifications/Email.php
```diff
@@ -0,0 +1,139 @@
...
+	 * Factory method.
+	 * @param type $userid
+	 * @param type $email
+	 * @param type $handle
+	 * @return \self
+	 * @throws FatalErrorException
+	 */
+	public static function CreateByUserID($userid, $email = '', $handle = '')
+	{
+		if (!(int)$userid) {
```
Mariusz08: PSR-2

qa-include/Q2A/Notifications/Mailer.php
```diff
@@ -0,0 +1,75 @@
...
+
+		if (qa_opt('smtp_active')) {
+			$this->isSMTP();
+			$this->Host = qa_opt('smtp_address');
+			$this->Port = qa_opt('smtp_port');
+
+			if (qa_opt('smtp_secure')) {
+				$this->SMTPSecure = qa_opt('smtp_secure');
+			} else {
+				$this->SMTPOptions = array(
```
Mariusz08: Short array - `[]`

svivian:
Hi @Spudley, apologies for the massive delay in getting to this. Finally found some time to test and merge. Thanks for the contribution!

FYI I went ahead and converted the classes to use our new namespace system (in the `qa-src/` directory).

PR #631, opened 2018-04-09
Fix another PHP 7.2 count warning

pupi1985:
Technically, the most appropriate check would be: `isset()`, `is_array()`, `!empty()`. However, as we know it is an array or it is not set we can remove the `is_array()` check. Furthermore, I think it is always set... but maybe I'm missing something.


svivian:
Thanks. Yeah `$favoritemap` is always an array but `$favoritemap['category']` may not be set. So `!empty` serves the same purpose here.

PR #637, opened 2018-05-18
chore: clone last 5 commits

DanielRuf:


svivian:
I guess the purpose of this is to make checkouts faster on Travis, is that right?

DanielRuf:
> I guess the purpose of this is to make checkouts faster on Travis, is that right?

You are right. If this is not the case and the benefits are very little I would close the PR.

PR #639, opened 2018-05-28
Fix count() notice on PHP 7.2

jairlopez:
It appears with a vanilla Q2A installation, after logging in as admin and clicking on admin -> My Updates


svivian:
Sorry for the delay. The bug fix commit has been merged. Thanks!

PR #650, opened 2018-08-05
Fix .qa-footer position always-visible fixed to bottom

alariva:
Make the .qa-footer of SnowFlat theme position always-visible fixed to bottom.

This makes some views look better specially when main content is less than viewport height.

Before change:

footer-before

After change:

footer-after

---
This change is [Reviewable](https://reviewable.io/reviews/q2a/question2answer/650)


alariva:
Thanks for pointing out your concerns @pupi1985 

However, it looks like you are against *any footer at all* instead of *sending footer to bottom*.
Footer is there, comes out of the box with the app. When the content does not fill the entire screen, this footer (with class `qa-footer`) steps in the **middle** of the screen, the rest of the screen is empty content. I don't see that to be *elegant*, do you?.

Headers belong to top, and footers belong to bottom. Most designers agree with that, do we agree on that?

This is my intent

I understand that fixed content may have issues. The best fit for *implementation* is something we may discuss.

In that sense, do you have any suggestion like other CSS code or is there a way to extend/override the theme CSS without loosing changes after an upgrade?

Thanks in advance.

PR #663, opened 2018-09-04
allow left, right and center aligned cells

fardina:


fardina:
This issue still exists in q2a verion 1.8.4. I am using markdown editor. In markdown we could have tables like:
| Syntax      | Description |
| ----------- | ----------- |
| Header      | Title       |
| Paragraph   | Text        |
and we could have left, right and center aligned text using : at the first, end or both side of hyphens. But at the moment all the tables will be left aligned.

PR #666, opened 2018-09-06
Remove some unnecessary SQL queries

pupi1985:
This is not just saving a few SQL queries but also saving some additional queries that addons may run.

Also I'm adding the `qa_db_query_sub_params()` function which is much more flexible than its sibling `qa_db_query_sub()`. I noticed that I ended up needing it in most of my plugins.

#666 :smiling_imp:


svivian:
Yes, the `qa_db_query_sub_params` is a good idea. Historically the "meta functions" like `func_get_args` have been slow - though the difference nowadays is probably small. But it does make sense to use more "direct" and obvious code where possible :)

PR #671, opened 2018-09-15
Fix closed questions by selected answer not displayed as closed

pupi1985:
It seems the core doesn't properly handle the `do_close_on_select` setting. It seems the rules were the only ones that were taking care of that setting:

https://github.com/q2a/question2answer/blob/1f39ae44adf42c83042be631e39373893a3ec969/qa-include/pages/question-view.php#L121

The rest of the core only considers a non-null `closedbyid` as the indicator of a closed question. I can think of two options to fix this:

1. Allow answer ids in the `closedbyid` so that the selected answer takes the place of the note or question duplicate. This approach seems to be the best one but is a bit non-backwards compatible as now questions can be closed by answers. Also SQL queries that consider closed questions as non-null `closedbyid` would be accurate (note they are not accurate currently).
1. Change the interpretation of closed questions along the core. This approach is more backwards compatible. Better for a workaround.

In either case, the selection of answers should fire the appropriate events in questions.

### DIFF
--- a/qa-include/ajax/answer.php
+++ b/qa-include/ajax/answer.php
@@ -37,7 +37,7 @@ list($question, $childposts) = qa_db_select_with_pending(
 
 // Check if the question exists, is not closed, and whether the user has permission to do this
 
-if (@$question['basetype'] == 'Q' && !isset($question['closedbyid']) && !qa_user_post_permit_error('permit_post_a', $question, QA_LIMIT_ANSWERS)) {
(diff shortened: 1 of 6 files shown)


svivian:
Thanks, all looks good. I guess we should be able to extract the closing-check logic to a function? Might make some parts a little simpler.

pupi1985:
Yes, I thought about it. I wasn't sure whether to add more stuff to the global scope or make the code slightly harder to read. Added it here #697

PR #672, opened 2018-09-19
Turn all remaining AJAX calls in admin into JSON

pupi1985:
This PR should be merged after #608. That would decrease the amount of changes in this PR.

Note this is a refactor in terms of output. The logic of the processes should remain the same.


svivian:
Fantastic work!! Huge thanks for doing this, and sorry for taking so long to get to it. I love that it updates the counts at the top of the Moderate/Hidden pages 😄

PR #685, opened 2018-11-05
Avoid warning when a user has no favorite

gturri:
When a user has no favorite (ie: when `$favoritemap` is an empty array) then displaying for instance the main page, generates this warning:

    Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/html/q2a/qa-include/app/format.php on line 384

This patch avoids filling my Apache error logs with those warnings


gturri:
I've just read the contribution guidelines (yeah, I should have read them beforehand) and saw that I should have pushed on the `bugfix` branch... and looking at this branch I just noticed that a similar patch has already been integrated with 6890815a4a445779696f7c5a8c040fa530ba2dfe . 
Sorry for the noise.

PR #689, opened 2018-11-17
Added checkmark to admin labels

q2apro:
Added checkmarks to strings that are showing up in admin/stats after successfully finished progress. This helps a lot to better see a finished progress,


svivian:
Thanks, nice addition.

Issue #706, opened 2018-12-08
Language strings in core plugins


svivian:
As noted [in this question](https://www.question2answer.org/qa/70221/why-dont-built-in-plugins-include-language-phrases), many of the default plugins in Q2A (the ones in `qa-plugin/`) do not use language phrases, meaning they are unable to be translated.

I'm opening this issue in case anyone wants to have a go at fixing it. As they are removable plugins the language files should follow [this guide](http://docs.question2answer.org/plugins/#plugin-localization). The current English phrases would go in a file like `qa-lang-default.php` in each plugin's directory.

Please target any PRs to the `dev` branch.

(To be clear, this is only about extracting the English phrases to the language files, not adding additional languages. Site owners using other languages would need to add their own files like `qa-lang-de.php`)

PR #707, opened 2018-12-09
Add localization to plugins under qa-plugin

jairlopez:
#706


qa-plugin/wysiwyg-editor/qa-plugin.php
```diff
@@ -42,5 +42,6 @@
 
 qa_register_plugin_module('editor', 'qa-wysiwyg-editor.php', 'qa_wysiwyg_editor', 'WYSIWYG Editor');
 qa_register_plugin_module('page', 'qa-wysiwyg-upload.php', 'qa_wysiwyg_upload', 'WYSIWYG Upload');
+qa_register_plugin_phrases('qa-wysiwyg-lang-*.php', 'wysiwyg');
```
pupi1985: Should the `'page'` modules be together followed by the `_phrases` or is my TOC kicking again? :)

pupi1985:
Looks much better to me :)

BTW, I've just realized that you don't add commas after the last element of an array. I've configured my IDE to add them automatically. I think Scott adds them too but maybe he can confirm that so that we avoid overwriting commas when formatting each other's code.

svivian:
Wow, kudos for the swift response! All looks good from a cursory glance on mobile, will merge in the next few days.

Re: arrays, yeah I usually do add trailing commas on multi-line arrays. I guess we don’t have a codesniffer rule for that implemented. Not a major issue though, I only spotted 2 instances anyway.

svivian:
Thanks again for this, good work!

Just an FYI, the language files should not have HTML in them. The way to use HTML is to pass that in with the substitution. I've made an update in da09b279e4d1e323f277b951607f9407c469fdf0 that does this.

jairlopez:
Awesome! Thank you for that.

PR #713, opened 2018-12-22
Added badges

ProThoughts:
added badges for number of issue, release and license. It will give quick info to users about q2a project.


ProThoughts:
Added info badges

PR #716, opened 2018-12-28
added more info to readme.md

ProThoughts:
Added more info to readme.md. this will help users to understand more about q2a project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,20 @@
-Question2Answer
------------------------------
- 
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for the suggestions but there were a few issues like repeating several of the bullet points. I've taken a few of the changes such as using the logo and headings.

ProThoughts:
Ok. I feel we should add Installation section. It will give idea to people that it takes less than 5 min to install q2a.

PR #721, opened 2018-12-30
Rm lang key "caching_dir_public"

gturri:
This key isn't used since f394a34c54b938218e8bb92549ae3a1f480e6f8a . Removing
it ensures translators don't bother translating it for nothing


svivian:
Thanks!

I did consider using that for a "warning" message instead of error but the current system doesn't allow for that, only errors. Plus seeing a warning all the time might get annoying.

PR #723, opened 2019-01-03
Fix selection of answers trying to close questions

pupi1985:
Steps to reproduce:
1. Disable `do_close_on_select`
2. Select an answer
3. The question is displayed as `reopened`

> If debugging is the process of removing bugs, then programming must be the process of putting them in


svivian:
I was just about to look into this today as members on my site reported the issue. Saved me a job, so thanks :)

PR #726, opened 2019-01-11
Fix file permissions

pupi1985:
Unrelated to this, I'm also thinking that there are a few issues with the `.gitignore` and the `.htaccess` files.

The `.gitignore` file should ignore the `.htaccess` file. The `.htaccess` file should be treated as _server-dependent_ in the same way as the `qa-config.php` file and be included as an example file.

Also the `.gitignore` for both the `.htaccess` and the `qa-config.php` file should also ignore only those two files that reside in that same directory. Currently it is ignoring all files with that name even if they are located in subdirectories.

If these changes make sense I can commit them to this branch.


svivian:
Thanks, and yes it would make sense to ignore htaccess, since people may customize it themselves and don't want to overwrite it with a new Q2A version.

PR #727, opened 2019-01-13
Remove .htaccess file from repository and ignore it in .gitignore

pupi1985:
Sending it to the bugfix branch because it shouldn't have any impact on currently existing installations


ProThoughts:
Nice fix! 

I'm expecting people will ask question in forum like how to make pretty url :)

pupi1985:
That shouldn't happen because I've already sent the pull requests to update the installation docs :)

ProThoughts:
👍

PR #729, opened 2019-01-15
Integration Q2A and phpBB

franzpc:
Maybe could you develop a plugin to integrate Q2A and phpBB.

I have a forum with phpBB, but I would like to migrate users to a new Q2A.
Q2A version: 1.8


pupi1985:
Please, don't create empty pull requests as a way of asking for features. Feature requests should go to the Q&A site. Needless to say, if you have written all the code that fulfills your requirement, go ahead and go for a pull request.

PR #731, opened 2019-01-18
Category add and edit event

thopd88:
I think it is necessary to have these events to make plugins be able to set category meta right after the category is created and/or for eventlog

### DIFF
--- a/qa-include/pages/admin/admin-categories.php
+++ b/qa-include/pages/admin/admin-categories.php
@@ -218,15 +218,34 @@ if (qa_clicked('docancel')) {
 					$recalc = $hassubcategory && $inslug !== $editcategory['tags'];
 				}
 
+				qa_report_event('cat_edit', qa_get_logged_in_userid(), qa_get_logged_in_handle(), qa_cookie_get(), array(
+					'categoryid' => $editcategory['categoryid'],
+					'parentid' => @$inparentid,
+					'content' => @$incontent,
+					'position' => @$inposition,
+					'name' => $inname,
+					'slug' => $inslug
+				));
+
 				qa_redirect(qa_request(), array('edit' => $editcategory['categoryid'], 'saved' => true, 'recalc' => (int)$recalc));
 
 			} else { // creating a new one
 				$categoryid = qa_db_category_create($inparentid, $inname, $inslug);
+				
 
 				qa_db_category_set_content($categoryid, $incontent);
 
 				if (isset($inposition))
 					qa_db_category_set_position($categoryid, $inposition);
+
+				qa_report_event('cat_new', qa_get_logged_in_userid(), qa_get_logged_in_handle(), qa_cookie_get(), array(
+					'categoryid' => $categoryid,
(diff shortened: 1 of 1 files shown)


qa-include/pages/admin/admin-categories.php
```diff
@@ -218,16 +218,35 @@
 					$recalc = $hassubcategory && $inslug !== $editcategory['tags'];
 				}
 
+				qa_report_event('cat_edit', qa_get_logged_in_userid(), qa_get_logged_in_handle(), qa_cookie_get(), array(
+					'categoryid' => $editcategory['categoryid'],
+					'parentid' => @$inparentid,
```
pupi1985: Any chance we can get rid of the `@`? It is just hiding errors. I'm not seeing this in context (only the added lines because I'm so lazy!) but checking with `empty()` or `isset()` is usually the classic `@`-removal pattern.

qa-include/pages/admin/admin-categories.php
```diff
@@ -218,16 +218,35 @@
...
+					'position' => @$inposition,
+					'name' => $inname,
+					'slug' => $inslug
+				));
+
 				qa_redirect(qa_request(), array('edit' => $editcategory['categoryid'], 'saved' => true, 'recalc' => (int)$recalc));
 
 			} else { // creating a new one
 				$categoryid = qa_db_category_create($inparentid, $inname, $inslug);
+
```
pupi1985: If the HTML in GitHub is not playing tricks on me, I see tabs in this line. It should be adding whitespace errors. I'd rather remove the line (2 empty lines as separators are IMO, a bit overkill).

qa-include/pages/admin/admin-categories.php
```diff
@@ -218,6 +218,15 @@
 					$recalc = $hassubcategory && $inslug !== $editcategory['tags'];
 				}
 
+				qa_report_event('cat_edit', qa_get_logged_in_userid(), qa_get_logged_in_handle(), qa_cookie_get(), array(
+					'categoryid' => $editcategory['categoryid'],
+					'parentid' => @$inparentid,
+					'content' => @$incontent,
+					'position' => @$inposition,
+					'name' => $inname,
+					'slug' => $inslug
```
pupi1985: There is a not-so-strong convention to use commas at the end of the array

qa-include/pages/admin/admin-categories.php
```diff
@@ -228,6 +237,15 @@
 				if (isset($inposition))
 					qa_db_category_set_position($categoryid, $inposition);
 
+				qa_report_event('cat_new', qa_get_logged_in_userid(), qa_get_logged_in_handle(), qa_cookie_get(), array(
```
pupi1985: Make sure to require `app/cookies.php` in order to safely use `qa_cookie_get()`

qa-include/pages/admin/admin-categories.php
thopd88: I agree, using `isset` is better than `@`

PR #738, opened 2019-01-31
Fix negative upvoteCount wrongly displayed

### LINKED ISSUE #733: Incorrect Schema.org's itemprop upvoteCount for negative score
Q2A version: 1.8.3

If a post has got a negative score, its itemprop upvoteCount used for Schema.org microdata will be incorrect and will contain an absolute value of the actual score. 

Eg. a post with score -2 results in the following HTML:

`<span class="qa-netvote-count-data">–2</span>
<span class="qa-netvote-count-pad"> votes 
<meta itemprop="upvoteCount" content="2">
</span>`

pupi1985:
Fixes #733

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -495,7 +495,7 @@ function qa_post_html_fields($post, $userid, $cookieid, $usershtml, $dummy, $opt
 
 		// schema.org microdata - vote display might be formatted (e.g. '2k') so we use meta tag for true count
 		if ($microdata) {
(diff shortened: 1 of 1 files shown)


svivian:
Bit of a strange one... I'm not sure exactly why we use `&ndash;` instead of a regular minus sign, if I'm not mistaken that's the wrong character anyway (should be used for ranges, not negativity). I don't see a need to overwrite `$netvotes` with its absolute value (we could use a separate variable for those `== 1` checks).

PR #739, opened 2019-02-01
Check the handle and return at the very beginning if it is null in

q2amarket:
The issue I came up when was working on one plugin to render handle html. The function was returning `Anonymous` when the handle is `NULL`

### DIFF
--- a/qa-include/app/users.php
+++ b/qa-include/app/users.php
@@ -560,6 +560,16 @@ if (QA_FINAL_EXTERNAL_USERS) {
 	{
 		if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
 
+        /**
+         * TODO: Q2A Market Fix: Check $handle and return at the very beginning if it is null.
+         * Currently in some scenario if handle is null than
+         * the function is rendering `Anonymous` text.
+         * This will prevent to render `anonymous` in many cases.
+         */
+		if (is_null($handle)) {
+            return;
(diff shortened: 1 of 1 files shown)


svivian:
What should it be displaying instead? If the username is NULL that means it is an anonymous user (or one that was deleted).

q2amarket:
> What should it be displaying instead? If the username is NULL that means it is an anonymous user (or one that was deleted).

Well, within the core it may make sense as it is calling only when the users is exists or not. However, in some cases it returns `Anonymous` when the data returns null as a handle value. For instance, I am using this `$handles  = qa_userids_to_handles($user_ids);`  and using `qa_get_one_user_html($handle)` in `foreach` loop returning `Anonymous` if array is null. 

For now I have unset the value in my own code but it would be better to check within the core function.

svivian:
Where are you getting the `$user_ids` from? Surely wherever you get those from should be fetching valid user IDs only. If that's not possible, then you can check in your foreach loop rather than unset, for example:

    $handles = qa_userids_to_handles($user_ids);
    foreach ($handles as $handle) {
        if ($handle === null) {
            $html = ''; // or whatever you need to do if null
        } else {
            $html = qa_get_one_user_html($handle);
        }
    }

I think `qa_get_one_user_html` is working as intended, changing it would break a bunch of other code that relies on displaying "anonymous".

q2amarket:
Okay let me try once again to check as you mentioned. Actually, I have a bit different scenario and it is creating an issue. I have to unset the element.

Let me have a look and get back to you.

PR #740, opened 2019-02-07
Add style element to qa_sanitize_html()

alacn1:
It's possible to inject style element when posting a question/answer and with it override page css.
This commit add style element to qa_sanitize_html().


svivian:
Wow, somehow I completely missed this PR. A belated thanks for the contribution!

PR #743, opened 2019-02-25
Panel of stats

franzpc:
Maybe you can develop a panel stats (like Google Analytics).


pupi1985:
@franzpc You're again not using GitHub properly. See [this comment](https://github.com/q2a/question2answer/pull/729#issuecomment-454569790) for more information.

PR #752, opened 2019-05-04
added robots.txt file.

ProThoughts:
This can be default robots.txt file for Q2A. User can disallow more crawlers based on their server log.

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,0 +1,67 @@
+# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+#
+User-agent: *
+Disallow: /login
+Disallow: /forgot
+Disallow: /search?
+
+# add Crawl delay
+Crawl-delay: 4
+
+User-agent: Yandex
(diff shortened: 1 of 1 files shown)


phy25:
Could you please explain why some specific bots needs to be blocked by default? I don't think it's wise to make them as built-in.

Also is it possible to use relative path for the sitemap?

ProThoughts:
@phy25 those specific bots I have seen on many q2a sites so added to file. Those are based on my experience. 

relative path are not possible, [see this thread. ](https://stackoverflow.com/questions/14196801/can-a-relative-sitemap-url-be-used-in-a-robots-txt)

svivian:
Thanks for the suggestion, however I don't think it's appropriate for us to dictate to users whether they should block certain bots. Also, adding this file will also cause it to overwrite whatever site owners already have, and prevents them from customizing it since it could get overwritten in future.

Feel free to put your suggestions as an answer on Q2A Q&A, if there is an appropriate question about robots.txt.

PR #755, opened 2019-05-11
Fire u_delete_before event before deleting a user

pupi1985:
This is useful whenever some action needs to be taken before deleting the user. For example, performing some action on their votes. Once the user is removed, it is not possible to access that information.

Changes to the events documentation have been added here: https://github.com/q2a/q2a.github.io/pull/25


q2apro:
Very good idea.

PR #764, opened 2019-06-27
Fix many failing SQL queries

pupi1985:
These are all related to the `qa_opt()` function returning strings and the new DB handling approach expects integers to actually be integers.

Although this could be fixed inside each of the functions, if the PHPDocs state a function receives an int, it should be an int. That's why I believe the casting should be done outside and, of course, to avoid adding so much _defensive code_.

### DIFF
--- a/qa-include/ajax/asktitle.php
+++ b/qa-include/ajax/asktitle.php
@@ -32,7 +32,7 @@ $doaskcheck = qa_opt('do_ask_check_qs');
 $doexampletags = qa_using_tags() && qa_opt('do_example_tags');
 
 if ($doaskcheck || $doexampletags) {
-	$countqs = max($doexampletags ? QA_DB_RETRIEVE_ASK_TAG_QS : 0, $doaskcheck ? qa_opt('page_size_ask_check_qs') : 0);
+	$countqs = max($doexampletags ? QA_DB_RETRIEVE_ASK_TAG_QS : 0, $doaskcheck ? (int)qa_opt('page_size_ask_check_qs') : 0);
 
 	$relatedquestions = qa_db_select_with_pending(
 		qa_db_search_posts_selectspec(null, qa_string_to_words($intitle), null, null, null, null, 0, false, $countqs)
@@ -56,7 +56,7 @@ if ($doexampletags) {
 	$exampletags = array();
 
 	$minweight = exp(qa_match_to_min_score(qa_opt('match_example_tags')));
-	$maxcount = qa_opt('page_size_ask_tags');
+	$maxcount = (int)qa_opt('page_size_ask_tags');
 
 	foreach ($tagweight as $tag => $weight) {
 		if ($weight < $minweight)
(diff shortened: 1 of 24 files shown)


svivian:
Where exactly are the queries failing? For example the user activity page loads fine for me without errors.

pupi1985:
I see this almost everywhere. This is what I get when I try to access the `/tags` page:
```
PHP Question2Answer MySQL query error 42000: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ''2') y ON qa_words.wordid=y.wordid' at line 1 - Query: SELECT word, tagcount FROM qa_words JOIN (SELECT wordid FROM qa_words WHERE tagcount>0 ORDER BY tagcount DESC LIMIT ?,?) y ON qa_words.wordid=y.wordid
```
Or trying to access my profile I get:
```
PHP Question2Answer MySQL query error 42000: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ''10'' at line 1 - Query: SELECT messageid, fromuserid, touserid, content, format, UNIX_TIMESTAMP(qa_messages.created) AS `created`, ufrom.flags AS `fromflags`, ufrom.level AS `fromlevel`, ufrom.email AS `fromemail`, ufrom.handle AS `fromhandle`, BINARY ufrom.avatarblobid AS `fromavatarblobid`, ufrom.avatarwidth AS `fromavatarwidth`, ufrom.avatarheight AS `fromavatarheight`, uto.flags AS `toflags`, uto.level AS `tolevel`, uto.email AS `toemail`, uto.handle AS `tohandle`, BINARY uto.avatarblobid AS `toavatarblobid`, uto.avatarwidth AS `toavatarwidth`, uto.avatarheight AS `toavatarheight` FROM qa_messages LEFT JOIN qa_users ufrom ON fromuserid=ufrom.userid LEFT JOIN qa_users uto ON touserid=uto.userid WHERE type='PUBLIC' AND touserid=(SELECT userid FROM qa_users WHERE handle=? LIMIT 1) ORDER BY qa_messages.created DESC LIMIT ?,?
```
If you have not faced these, then we must have some kind of setting different in our environments.

svivian:
I'm not getting those errors. It seems to be due to the LIMIT clause, which I thought I'd fixed in b8d3758.

Do you have [this line](https://github.com/q2a/question2answer/blob/dev/qa-src/Database/DbConnection.php#L221) in your code? That should bind the parameters as integers so they're not quoted. I wonder if it's a MariaDB issue.

pupi1985:
I see. It seems the issue is that you have to cast the parameter to INT when it is an integer:
```php
if (filter_var($param, FILTER_VALIDATE_INT) !== false) {
    $dataType = PDO::PARAM_INT;
    $param = (int)$param;
} else {
    $dataType = PDO::PARAM_STR;
}
```

After doing so the queries started to work again. So maybe that was it.

PR #772, opened 2019-09-06
Get proper client IP. Checking proxy instead

hrickmachado:
I am checking if the HTTP_X_FORWARDED_FOR to get the client ip.
https://en.wikipedia.org/wiki/X-Forwarded-For

My Q2A did not count the number of visits properly with proxy enabled.

### DIFF
--- a/qa-include/qa-base.php
+++ b/qa-include/qa-base.php
@@ -1260,8 +1260,17 @@ function qa_clicked($name)
 function qa_remote_ip_address()
 {
 	if (qa_to_override(__FUNCTION__)) { $args=func_get_args(); return qa_call_override(__FUNCTION__, $args); }
-
-	return isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : null;
+	
+	if (array_key_exists('HTTP_X_FORWARDED_FOR', $_SERVER)) {
+		$parsedUrl = parse_url($_SERVER["HTTP_X_FORWARDED_FOR"]);
+		return $parsedUrl['host'];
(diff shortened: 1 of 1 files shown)


pupi1985:
It's not a big deal as Q2A only uses IP addresses to perform rate limits or maybe to block them. I mean, you won't be able to access admin settings if you spoof your IP address like that. However, due to the existence of plugins, it is fair to take that into account, particularly, if it is just a checkbox that could fit in the admin/general section.

hrickmachado:
How to make it configurable with a default of off ? 
Where is the configuration implemented ?

> This may cause issues if the server did not has traffic go through a proxy, and malicious clients requests server with fake X-FORWARDED-FOR HTTP header. Please at least make it configurable with a default of off.

svivian:
Hi, thanks for your contribution. However, as phy25 states this is insecure as it can be easily spoofed. The correct solution, if you wanted to do this in Q2A, is when the X-Forarded-For exists, check that `$_SERVER['REMOTE_ADDR']` matches the approved IP addresses of your proxy.

The other (and often easier) solution is to handle that at the server level. [Here are some examples using CloudFlare](https://support.cloudflare.com/hc/en-us/articles/200170786-Restoring-original-visitor-IPs-Logging-visitor-IP-addresses-with-mod-cloudflare-) you can modify for your own use.

As doing this within Q2A itself is somewhat niche I'd suggest creating a [function override plugin](https://docs.question2answer.org/plugins/overrides/) for `qa_remote_ip_address`. That way you can handle the correct IPs for your own system.

PR #784, opened 2019-11-22
Add docker, docker stack and helm deployment support

jloaiza:
This commit provides a Dockerfile based on php-fpm that contains the source
code of question2answer. The container also support init scripts that may be
used to install plugins, themes or configure the running instance.

The commit also provides a compose file which enables the possibility to run a
deploy of question2answers using docker stack.

Also includes a helm directory which enables the deploy of question2answers in
a kubernetes server using helm. For this, a public image of the Dockerfile is
needed but for now, there is none published; so you may have publish your own
until an official one is released.

Documentation about this files, how to deploy an image, a docker stack or helm
instance is also provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,15 @@ Official releases can be found either on the [Q2A website][Q2A] or via the Relea
 [6]: https://docs.question2answer.org/themes/
 [7]: https://docs.question2answer.org/translate/
 [8]: https://docs.question2answer.org/addons/
+
+## Deployment
+
+Docker, Docker Stack and Kubernetes
+-----------------------------------
+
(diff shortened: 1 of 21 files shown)


svivian:
Hi, thanks for the PR. Unfortunately this kind of thing is not appropriate for the Q2A core repository. If you have a well-maintained docker image (e.g. on docker hub) I'd be happy to link to it from [the docs](https://github.com/q2a/q2a.github.io/). If it needs a huge amount of setup explanation, as suggested by all the PR files, then you could create a blog post somewhere and I'll link that too.

binarious:
Too bad this wasn't accepted. Thanks for the work, @jloaiza! This helped us a lot.

PR #798, opened 2020-02-25
Added support for utf8mb4 (emoji support)

bertrandgorge:
Fixes issue #791

### DIFF
--- a/qa-include/db/install.php
+++ b/qa-include/db/install.php
@@ -67,6 +67,11 @@ function qa_db_table_definitions()
 	require_once QA_INCLUDE_DIR . 'db/maxima.php';
 	require_once QA_INCLUDE_DIR . 'app/users.php';
 
+	if (defined('QA_USE_UTF8MB4') && QA_USE_UTF8MB4)
+		$collation = 'utf8mb4_bin';
+	else
+		$collation = 'utf_bin';
+
 	/*
 		Important note on character encoding in database and PHP connection to MySQL
 
@@ -107,7 +112,7 @@ function qa_db_table_definitions()
 			'avatarheight' => 'SMALLINT UNSIGNED', // pixel height of stored avatar
 			'passsalt' => 'BINARY(16)', // salt used to calculate passcheck - null if no password set for direct login
 			'passcheck' => 'BINARY(20)', // checksum from password and passsalt - null if no password set for direct login
-			'passhash' => 'VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_bin DEFAULT NULL', // password_hash
+			'passhash' => 'VARCHAR(255) CHARACTER SET utf8 COLLATE '.$collation.' DEFAULT NULL', // password_hash
 			'level' => 'TINYINT UNSIGNED NOT NULL', // basic, editor, admin, etc...
 			'loggedin' => 'DATETIME NOT NULL', // time of last login
 			'loginip' => 'VARBINARY(16) NOT NULL', // INET6_ATON of IP address of last login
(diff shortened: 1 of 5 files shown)


qa-include/db/install.php
```diff
@@ -67,6 +67,11 @@ function qa_db_table_definitions()
 	require_once QA_INCLUDE_DIR . 'db/maxima.php';
 	require_once QA_INCLUDE_DIR . 'app/users.php';
 
+	if (defined('QA_USE_UTF8MB4') && QA_USE_UTF8MB4)
```
svivian, line 70: As this same check is used many times, I'd suggest extracting it into a function e.g. `qa_db_get_mysql_collation()`

svivian:
Hi, sorry for the late reply. Thanks for the PR, from a first glance it looks great! Would you be able to make the PR against the `dev` branch, as that is the one for the next major version?

At the moment it appears we can only set utf8mb4 prior to starting a new Q2A instance. One thing to consider is whether to add an upgrade path for existing sites. However it may be complex, as we need to detect the MySQL version and only upgrade those on 5.5.3+. But then the upgrade would need to be done on future versions in case they upgraded their MySQL version later.

bertrandgorge:
Hello Scott, yes, I thought about the upgrade path, and I believe it's a lot of work for something people should do manually if they wish to. If you want I can make a "how to upgrade" for people who'd want to do that.

As for the dev branch, I'm not sure I know how to do that ? Should I create another PR ?

rklec:
So is there any news here? Emoji support is really needed _by default_ for q2a.

svivian:
@rklec Yes I plan on merging it when I get a free moment.

qa-include/db/install.php
```diff
@@ -107,7 +112,7 @@ function qa_db_table_definitions()
 			'avatarheight' => 'SMALLINT UNSIGNED', // pixel height of stored avatar
 			'passsalt' => 'BINARY(16)', // salt used to calculate passcheck - null if no password set for direct login
 			'passcheck' => 'BINARY(20)', // checksum from password and passsalt - null if no password set for direct login
-			'passhash' => 'VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_bin DEFAULT NULL', // password_hash
+			'passhash' => 'VARCHAR(255) CHARACTER SET utf8 COLLATE '.$collation.' DEFAULT NULL', // password_hash
```
rklec, line 115: You also need to change the charset if you change the collation, apparently...

Otherwise you may get an error like:
> COLLATION 'utf8_bin' is not valid for CHARACTER SET 'utf8mb4'

(At least I got it in MariaDB)

```suggestion
			'passhash' => 'VARCHAR(255) CHARACTER SET utf8mb4 COLLATE '.$collation.' DEFAULT NULL', // password_hash
```

(Of course may be adjusted to work with the defined `QA_USE_UTF8MB4` option.)

rklec (changes requested):

svivian:
Hi @bertrandgorge sorry I never got around to answering you. If you're able to create a new PR on the dev branch that would be great, thanks. Not sure how easy it would be... I think you may need to check out a fresh copy of the dev branch and re-apply the changes. (If that's too much hassle let me know and I'll try and sort it.)

rklec:
Actually, you just need to click "edit" at the top right of this PR and change the branch you want to merge this in. Depending on the changes you may of course need to merge in the dev branch afterwards.

bertrandgorge:
Hello @rklec I don't see that option.... @svivian if you can manage it, it would be great as I don't work on Question2Answer anymore unfortunately.

rklec:
You should be able to click "edit" and then change the branch below the title after the text "1 commit into".

bertrandgorge:
Got it. There's a conflict of course, coming from a refactorisation of qa-db.php...

svivian:
Yeah I’ve made a lot of changes in 1.9. Don’t worry about it, I’ll try and cherry pick what I can.

PR #803, opened 2020-02-29
use curl instead of file_get_contents

aben1188:
Due to the design of the file_get_contents function, sometimes getting external content will be very slow. You should use curl instead if possible. For details, see here: https://stackoverflow.com/questions/3629504/php-file-get-contents-very- slow-when-using-full-url

### DIFF
--- a/qa-include/qa-base.php
+++ b/qa-include/qa-base.php
@@ -1840,15 +1840,22 @@ function qa_retrieve_url($url)
 		return '';
 	}
 
-	$contents = @file_get_contents($url);
+        $contents = '';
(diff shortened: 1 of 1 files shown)


svivian:
Hi, thanks for the contribution! Would you be able to fix the Travis issues? Appears to be indenting issues. You can check for issues yourself locally with [PHPCS](https://github.com/squizlabs/PHP_CodeSniffer) using the command:

    php /path/to/phpcs.phar --extensions=php --standard=qa-tests/phpcs/ruleset.xml .

PR #807, opened 2020-05-06
Prevent using <embed> and <object> tags

gturri:


svivian:
Thanks for your contribution!

PR #814, opened 2020-05-14
Fix access array offset notice on post page.

springmaple:
Getting "Notice: Trying to access array offset on value of type null in C:\...\qa-include\pages\questino-view.php on line 1131" on version 1.8.4


svivian:
Under what circumstances did you get the notice? I don't see any notice when I view a question.

springmaple:
@svivian 
image

It appears on every question page, with or without logged in.

I am using PHP 7.4.1.

svivian:
Sorry I forgot about this PR. I actually fixed this notice, along with several others, a few weeks ago. Thanks for taking the time to report and make a PR.

PR #816, opened 2020-05-18
Accessibility improvements

klues:
Various accessibility improvements, e.g.:
* correct usage of semantic HTML5 elements (e.g. `main`, `nav`)
* correct relationships between input elements and labels (via properties `id` and `for`)
* added labels to non-labelled elements

**Note:** I've just extracted the most important changes from our Q2A instance which can be found at https://github.com/wbt-vienna/question2answer

**Important:** I've **not tested** this with all the standard Q2A Themes, most likely this PR will break some things. Possibly it makes sense to extract smaller changes and PRs and to not merge this big PR.

We're using Question2Answer on [q2a.asterics-foundation.org](https://q2a.asterics-foundation.org/questions) with [Donut theme](https://github.com/amiyasahu/Donut). There I've also made some changes regarding accessibility which can be found in [this PR](https://github.com/amiyasahu/Donut/pull/119).


svivian:
Haven't had a chance to review this properly yet, but just wanted to say this is a great area for improvement that we have neglected in the past. Thanks for the effort!

The branch seems to have gotten a bit messed up, but I should be able to cherry-pick your commits when I get time.

klues:
Yeah, as said, I just tried to extract all the relevant things from our [own project](https://github.com/wbt-vienna/question2answer) and made a PR out of it.

Just ask if any changes are unclear. It's for sure a good idea to split this PR into smaller ones with definite topics like e.g. correct usage of semantic HTML5 elements.

PR #830, opened 2020-07-20
Increase security by setting the httpOnly flag to session cookie

gturri:
According to Mozilla best-practices (see https://observatory.mozilla.org
and https://infosec.mozilla.org/guidelines/web_security#cookies )
session cookies should be created with the HttpOnly flag. This prevents
XSS vulnerabilities.

This PHPSESSID cookie is still sent over ajax calls (it just prevents JS to
access it) so it does not break any usage


svivian:
Thanks!

PR #833, opened 2020-08-01
qa_sanitize__html_hook_tag don't need to handle <embed> anymore

gturri:
Those two lines were used to remove potential "allowscriptaccess"
attribute from <embed> tag. It was probably made for security reasons.
But it turned out that even with this, it was dangerous to allow <embed>
tag. That's why 188f5e98792f2aca3afdd8a03d88e5732c7f5685 removed
completely the possibility to have <embed> tags (if there are some in
inputs, they are going to be completely deleted by htmLawed.
Consequently, the two lines removed by this commit are completely
useless since this other commit. Removing them makes this part of the
code a bit less complex to understand.

### DIFF
--- a/qa-include/qa-base.php
+++ b/qa-include/qa-base.php
@@ -1048,9 +1048,6 @@ function qa_sanitize_html_hook_tag($element, $attributes = null)
 	if ($element == 'param' && trim(strtolower(@$attributes['name'])) == 'allowscriptaccess')
 		$attributes['name'] = 'allowscriptaccess_denied';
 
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for the PR, but I think it's best to keep this just in case someone overrides the `qa_sanitize_html` and allows embed tags again. Even if it's not perfectly secure it does mitigate one potential attack.

gturri:
ah, you're right: I forgot that we can override methods and that this PR could impact plugins. Good catch.

PR #887, opened 2021-03-24
Bringing Composer to Q2A

renaatdemuynck:
Another try at bringing Composer to Q2A (previously done by @rampmaster in 2014)


svivian:
Hi, thanks for the great effort - this looks good! I think we are still not quite ready for Composer as I need to restructure the project (for example the vendor folder should not be public). I expected to be a lot further down that road by now but I don't have much free time so development has been slow recently. But I'll keep this around for a future Q2A version :)

PR #890, opened 2021-03-29
Db connection query failure params

Chepheus:
Hello, sorry for my English
I need this params to display whats wrong was in query.

### DIFF
--- a/qa-src/Database/DbConnection.php
+++ b/qa-src/Database/DbConnection.php
@@ -150,7 +150,7 @@ class DbConnection
 	 * @param int $errno
 	 * @param string $error
 	 * @param string $query
-     * @param array|null $params
+    * @param array|null $params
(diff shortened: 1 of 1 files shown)


svivian:
Thanks for the PR. Could you explain a little more what this is for? The `$params` variable is not being used in your PR.

Chepheus:
It is need for more detailed logs if inserting was failed.
I create plugin for import array of data and I don't want to stop importing if one of element is broken (by default script stops).
I need this params to understand whats wrong with my data without deep debugging.

Chepheus:
About `$params` - I override default function to `function qa_page_db_fail_handler($type, $errno = null, $error = null, $query = null, **array $params = null**)`

Chepheus:
What do you think about this PR? =)

PR #891, opened 2021-03-30
Add capability to provide custom date when create post

Chepheus:
Hello, sorry for my English.
I need this functionality to provide my custom date because of importing of data with predefining data.

### DIFF
--- a/qa-include/app/post-create.php
+++ b/qa-include/app/post-create.php
@@ -63,16 +63,17 @@ function qa_combine_notify_email($userid, $notify, $email)
  * @param string|null $extravalue
  * @param bool $queued
  * @param string|null $name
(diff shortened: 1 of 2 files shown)


svivian:
Hi, I think a simpler solution is to add your post without the date, then update it manually. `qa_post_create` returns the ID of the new post so you can run a new query  like`UPDATE ^posts SET created=$ WHERE id=#`

Chepheus:
I see, but I have a lot of data for question creating (about 4000+ questions) and insert one row and then update it. I think it's not optimal for my purpose.

Chepheus:
What do you think about this PR? =)

nguyentrinhan-dev (approved):

PR #894, opened 2021-05-02
Fix many issues regarding $errors variable and @ character

### LINKED ISSUE #888: PHP 8.0 Fatal Error
Doing some quick testing using PHP 8.0, I found a fatal error when visiting the "Register" page:

`2021/03/28 17:05:17 [error] 36104#36104: *73759 FastCGI sent in stderr: "PHP message: PHP Fatal error: Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in /var/www/example.com/htdocs/support/qa-include/app/captcha.php:98 Stack trace:`
`#0 /var/www/example.com/htdocs/support/qa-include/pages/register.php(219): qa_set_up_captcha_field()`
`#1 /var/www/example.com/htdocs/support/qa-include/app/page.php(183): require('...')`
`#2 /var/www/example.com/htdocs/support/qa-include/qa-page.php(47): qa_get_request_content()`
`#3 /var/www/example.com/htdocs/support/qa-include/qa-index.php(188): require('...')`
`#4 /var/www/example.com/htdocs/support/index.php(27): require('...')`
`#5 {main}`
`thrown in /var/www/example.com/htdocs/support/qa-include/app/captcha.php on line 98" while reading response header from upstream, client: 2805:1015:f:270e:c1e9:8a26:be3f:71bd, server: example.com, request: "GET /support/register?to= HTTP/2.0", upstream: "fastcgi://127.0.0.1:9000", host: "example.com", referrer: "https://example.com/support/"`

Would be nice if a warning message is added to the official installation page, (PHP 5.2 or later is recommended).
Normally, I use 7.4 for Q2A and everything runs just fine!

pupi1985:
I've removed many usages of the @ sign and initialized the `@errors` variable in all pages I noticed it was used.

I have not removed all @ signs, but this should get the code one step closer to that.

This PR should also fix #888.

### DIFF
--- a/qa-include/app/captcha.php
+++ b/qa-include/app/captcha.php
@@ -82,20 +82,22 @@ function qa_set_up_captcha_field(&$qa_content, &$fields, $errors, $note = null)
 	// workaround for reCAPTCHA, to load multiple instances via JS
 	$count = @++$qa_content['qa_captcha_count'];
 
+	$hasCapatchaError = array_key_exists('captcha', $errors);
+
 	if ($count > 1) {
 		// use blank captcha in order to load via JS
 		$html = '';
 	} else {
 		// first captcha is always loaded explicitly
 		$qa_content['script_var']['qa_captcha_in'] = 'qa_captcha_div_1';
-		$html = $captcha->form_html($qa_content, @$errors['captcha']);
+		$html = $captcha->form_html($qa_content, $hasCapatchaError ? $errors['captcha'] : null);
 	}
 
 	$fields['captcha'] = array(
 		'type' => 'custom',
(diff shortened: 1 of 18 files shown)


svivian:
Thanks for this. We ought to move to PHP 7 minimum soon, so we can write `$errors['key'] ?? null` instead of all the `isset()` calls. I've done that in other projects and it makes things a lot neater!

pupi1985:
I know, I'm looking forward for that to happen :) The PHP 8 release can be used as an excuse for this dependency update.

The good thing of having the wordy `isset()`s in the code (compared to the `@` sign) is that IDE's should be able to detect the "code smell" and hint the fix.

PR #905, opened 2021-07-16
Added alt text for user's Avatar

MominRaza:
Ex.: Momin's avatar
#902


phy25 (changes requested):
Alt text should be translatable

PR #911, opened 2021-08-25
Fix #898: error while adding categories in PHP 8

### LINKED ISSUE #898: Unable to add category as superadmin
I made a site by questions2answer.

But, when I try to add new category as superadmin, I receive HTTP ERROR 500.

I try all alternatives but I can't solve this problem.

What should I do?

thanks in advance.

pupi1985:
Fix #898 . More `@` operator issues

### DIFF
--- a/qa-include/pages/admin/admin-categories.php
+++ b/qa-include/pages/admin/admin-categories.php
@@ -480,7 +480,8 @@ if ($setmissing) {
 
 		else {
 			$positionvalue = isset($previous) ? qa_lang_html_sub('admin/after_x', qa_html($previous['title'])) : qa_lang_html('admin/first');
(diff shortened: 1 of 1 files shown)


tillkuhn:
Many thanks, @pupi1985, your patch worked for me! Are there any plans when this will be merged and released? Categories should be considered an essential feature. And if clients start a fresh installation based on PHP8 (which is not that unlikely since it's out for over a year now), it's completely unusable unless you find this issue and manually apply the changes on the server side.

pupi1985:
Sorry, but that's not really my call. I have the same privileges as you :)

tillkuhn:
@pupi1985 That's what I expected, but there was nobody else in this thread to address :-) Anyway, when looking at your recent PR activities, the core team should definitely promote you. Have a nice day!

PR #914, opened 2021-09-22
Fix several RTL issues in question views

pupi1985:


qa-theme/SnowFlat/qa-styles-rtl.css
```diff
@@ -310,8 +311,7 @@ h1 {
 }
 
 .qa-q-view-avatar {
-	right: auto;
-	left: 20px;
```
phy25, line 314: Any reason dropping the 20px here rather than keeping it?

phy25 (changes requested):

qa-theme/SnowFlat/qa-styles-rtl.css
pupi1985: I'd answer with the opposite question: what was the reason to add it there in the first place?

PR #928, opened 2021-12-10
display dohideall button only if the user has visible posts

ihlassovbetov:


svivian:
Hey, sorry for the late reply. Is there a particular use case for this? I think it's useful to keep the button always there for consistency.

PR #937, opened 2022-02-26
Fix PHP8 bugs

trentwiles:
Solves "Uncaught TypeError: array_key_exists()" when captcha is enabled.


yerbol89:
@pupi1985 I'm use PHP 7.4.30 version. And i see error 
`Warning: array_key_exists() expects parameter 2 to be array, null given in /var/www/stud.baribar.kz/data/www/surak.baribar.kz/qa-include/app/captcha.php on line 85`

pupi1985:
I believe I've fixed this in my other branch: https://github.com/pupi1985/question2answer/commits/bugfix-pupi1985

PR #952, opened 2022-07-22
Allow insecure requests in file_get_contents()

pupi1985:


svivian:
I don't think this is a good idea. If people have trouble with fetching secure URLs they need to update their server rather than make insecure requests and leave themselves open to an attack. If absolutely necessary, the function can be overridden in a plugin to be insecure.

PR #956, opened 2022-08-15
Fix wrong redirection when titles do not match

pupi1985:
Take this URL: `https://example.com/123/nice-title?show=29#c29` . Let's assume it works as excpected.
Now turn it into `https://example.com/123/?show=29#c29` . The result is taking away the comment reference.

We need to rebuild the URL.

NOTE: This is still leaving aside other states like editing a question. This should be less relevant.


svivian:
Thanks for bringing up the issue. I actually fixed this in a much simpler way: 9a0d1175e398315442a406df9161cedead9848c2 
That preserves any other parameters that may be part of the URL. (Browsers preserve the # when redirecting so we don't need to specify that)

PR #959, opened 2022-08-20
Main Navigation Reorder / Footer Scripts / Plugins' CSS

heliochun:
Re ordered "Ask a question" and "Admin" links in Main Navigation, so they're displayed at the very top in vertical menus (mobile devices for example).

For horizontal menus (Desktop) a 'floating:right;' can be applied for these menu items, so they're displayed at the far right end, as they were before.

### DIFF
--- a/qa-include/app/format.php
+++ b/qa-include/app/format.php
@@ -1461,7 +1461,7 @@ function qa_user_sub_navigation($handle, $selected, $ismyuser = false)
 {
 	$navigation = array(
 		'profile' => array(
-			'label' => qa_lang_html_sub('profile/user_x', qa_html($handle)),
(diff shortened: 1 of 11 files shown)


svivian:
This seems to be mixed in with some unrelated changes. I'm not sure there's a reason to move the navigation items, the ordering seems fine to me. Thanks though.

heliochun:
Yeah that part can be removed. I thought they would split into different pull requests but got bundled into one together.
It makes sense for easier access in vertical menus, like mobile devices, but I guess people can make this reorder on their themes.
I've checked the "Allow edits by maintainers" option, but idk if it'll work to just delete that commit/part.

Cheers

PR #968, opened 2022-10-10
Fix: Fix transparent background of top-right menu

emranbm:
# What's wrong?
Using the default theme (*SnowFlat*), the top right menu has a corrupted background, as shown in the image below.  
 
Screenshot from 2022-10-10 21-14-41

# What does this PR do?
This PR simply adds extra space for the menu container by extending its height. It will look like this:
Screenshot from 2022-10-10 21-14-10

### DIFF
--- a/qa-include/qa-theme-base.php
+++ b/qa-include/qa-theme-base.php
@@ -774,7 +774,7 @@ class qa_html_theme_base
 			if (isset($favorite))
 				$this->output('<form ' . $favorite['form_tags'] . '>');
 
-			$this->output('<div class="qa-main-heading">');
+			$this->output('<div class="qa-main-heading" dir="auto">');
 			$this->favorite();
 			$this->output('<h1>');
 			$this->title();
@@ -1753,7 +1753,7 @@ class qa_html_theme_base
 	public function q_item_title($q_item)
 	{
 		$this->output(
-			'<div class="qa-q-item-title">',
+			'<div class="qa-q-item-title" dir="auto">',
 			'<a href="' . $q_item['url'] . '">' . $q_item['title'] . '</a>',
 			// add closed note in title
 			empty($q_item['closed']['state']) ? '' : ' [' . $q_item['closed']['state'] . ']',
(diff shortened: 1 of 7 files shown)


pupi1985:
Can you provide the steps to replicate this issue?
Please, test in this default setup: https://www.question2answer.org/qa/

emranbm:
I've no idea why it's correct on the demo site. However, here's the reproducing environment:
Dockerfile:
```Dockerfile
FROM php:7.4.30-apache
RUN apt-get update && apt-get install -y curl libcurl4-openssl-dev
RUN docker-php-ext-install mysqli
RUN docker-php-ext-install curl
RUN docker-php-ext-install json
RUN apt-get update && apt-get install -y git
RUN git clone -b v1.8.6 --depth 1 https://github.com/q2a/question2answer.git /var/www/html/
```
docker-compose.yml:
```yaml
services:
  q2a:
    build:
      context: .
    depends_on:
      - database
    ports:
      - 80:80
    environment:
      DB_HOSTNAME: database
      DB_USERNAME: q2a
      DB_PASSWORD: dummy
      DB_DATABASE: q2a_db
  database:
    image: mysql:5.7.39
    environment:
      MYSQL_ROOT_PASSWORD: dummy
      MYSQL_DATABASE: q2a_db
      MYSQL_USER: q2a
      MYSQL_PASSWORD: dummy
```

pupi1985:
I'm unable to reproduce this in Google Chrome 106.0.5249.91 (Official Build) (64-bit) and Mozilla Firefox 102.3.0esr. What browser are you using?

Side notes:
 * You can use php:7.4-apache and that will pull PHP 7.4.32, which has the latest bugfixes
 * `curl` is already part of the image. No need to reinstall it
 * `curl` and `json` shouldn't require any PHP setup
 * Try using the `bugfix` branch instead of the `v1.8.6` tag
 * Clear your browsers cache or test this in an incognito tab

emranbm:
Thanks for the side notes. I applied them, but no luck. Encountered the corrupted menu on two different machines, persistently.  
Chrome version: 105.0.5195.102
Firefox version: 105.0

Have you tried the above-mentioned environment? It should reproduce the issue.

pupi1985:
I tested your environment and it works perfectly fine. Considering this version of Q2A has been around for 1.5 years, the steps to reproduce the issue don't reproduce it, and nobody has mentioned this so far, I have to assume there is something wrong with your setup (operating system, browser, etc.). The fact that this is replicated on another system, makes me wonder if it has been improperly setup as well.

emranbm:
Ok, I've patched it locally for my environment.  
However, didn't get why it's not happening in your system.

PR #983, opened 2023-06-09
New version and some new features

PCcloob:
Hi ,
When will the new version be released?
I have suggested changes to the site that need to be done.
When a user sends a post, he waits for the administrator's approval and the administrator must approve his post.
Here, facilities should be created so that if the post needs to be edited, a command will be sent under the post to the user who can edit his own post and if it is finalized, the administrator will publish it.
And the user can see his published and unpublished posts on a page, and also see the reason why his posts are not published, so that he can correct them for the administrator to publish.
The second thing is that it is possible to log in through a mobile phone so that the user can enter the site without entering a password and using a mobile phone and a OTP code.


svivian:
This is not a pull request...

The new version will be released when I get some time. Meanwhile you can use the latest version from the "bugfix" branch, it has the fixes for PHP 8.

PR #986, opened 2023-08-16
Inscreased HTML wrapper scope by 1, to better handle UI components/Styling

rxchun:
Basically just moved `qam-account-items-wrapper` outside `nav_user_search()` , instead of inside.
This gives plugins the ability to output stuff right next to the User Avatar on Modern themes such as SnowFlat.
For example, the OnSiteNotification bell - https://i.ibb.co/9VMxz7z/z317.jpg

```
$this->output('<div class="qam-account-items-wrapper">');
	$this->nav_user_search();
$this->output('</div>');
```

### DIFF
--- a/qa-theme/SnowFlat/qa-theme.php
+++ b/qa-theme/SnowFlat/qa-theme.php
@@ -177,8 +177,6 @@ class qa_html_theme extends qa_html_theme_base
 	public function nav_user_search()
 	{
 		// outputs login form if user not logged in
-		$this->output('<div class="qam-account-items-wrapper">');
-
 		$this->qam_user_account();
 
 		$this->output('<div class="qam-account-items clearfix">');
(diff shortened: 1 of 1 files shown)


svivian:
I don't understand what this solves, as the div already wraps everything inside the `nav_user_search()` method.

rxchun:
Because the header structure is like this:

```
public function header()
{
$this->output('<div class="qa-header">');

    $this->logo();
    $this->nav_user_search();
    $this->nav_main_sub();
    $this->header_clear();

    $this->output('</div> <!-- END qa-header -->', '');
}
```

So if you append/prepend something to the `nav_user_search()` like this:

```
function nav_user_search() {
    $this->output('<!-- Output something here -->');
    qa_html_theme_base::nav_user_search();
}
```

it will append outside the `qam-account-items-wrapper` , so it will be outputted as a sibling of the `logo()`, `nav_user_search()` and `nav_main_sub()` . But if you wrap it outside, you can properly append stuff as siblings of the User Avatar and prevent this: https://question2answer.org/qa/111662/notification-moved-to-left-please-help

rxchun:
So that the content gets properly divided by sections like this

PR #996, opened 2024-05-22
Related Question Stats - Sidebar

rxchun:
This adds an Answer Counter to the Related Questions widget, on the sidebar. 
Screenshot: https://i.ibb.co/vsYqgtZ/z496.jpg

If people don't want to display stats, they can just comment out the new added line in:
qa-include/plugins/qa-widget-related-qs.php

Line to comment out:
https://github.com/q2a/question2answer/commit/07bcf491e4b125dcc91527813f7b3b3479918fba#diff-49b7248c15fca491a6271afeff134615cc8323afbe63706402ae6369ea32c6adR73

### DIFF
--- a/qa-include/plugins/qa-widget-related-qs.php
+++ b/qa-include/plugins/qa-widget-related-qs.php
@@ -65,8 +65,12 @@ class qa_related_qs
 			$themeobject->output('<ul class="qa-related-q-list">');
 
 			foreach ($questions as $question) {
+				$isASelected = $question['selchildid'] != null ? ' qa-a-count-selected' : '';
+				$qaLangAnswers = str_replace(':', '', qa_lang_html('profile/answers'));
+				
 				$themeobject->output(
 					'<li class="qa-related-q-item">' .
+					'<span class="qa-related-q-item-stats'. $isASelected .'" title="'.$qaLangAnswers.'">' . qa_html($question['acount']) . '</span>' .
 					'<a href="' . qa_q_path_html($question['postid'], $question['title']) . '">' .
 					qa_html($question['title']) .
 					'</a>' .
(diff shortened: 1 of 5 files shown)


qa-include/plugins/qa-widget-related-qs.php
```diff
@@ -65,8 +65,12 @@ public function output_widget($region, $place, $themeobject, $template, $request
 			$themeobject->output('<ul class="qa-related-q-list">');
 
 			foreach ($questions as $question) {
+				$isASelected = $question['selchildid'] != null ? ' qa-a-count-selected' : '';
```
pupi1985, line 68: `$isASelected` suggests a boolean variable. Wouldn't it make more sense like `$selectedClass` or something similar?

As a side note, isn't an affirmation easier to read than a negation? E.G.: something like this `isset($question['selchildid']) ? ' qa-a-count-selected' : '';`

qa-include/plugins/qa-widget-related-qs.php
```diff
@@ -65,8 +65,12 @@ public function output_widget($region, $place, $themeobject, $template, $request
...
+				$qaLangAnswers = str_replace(':', '', qa_lang_html('profile/answers'));
```
pupi1985, line 69: I wouldn't manipulate or make any assumption on user input. Languages are very different from each other. I'd suggest to use this instead: `main/x_answers`

qa-theme/Candy/qa-styles.css
```diff
@@ -453,7 +453,27 @@ h2 {font-size:22px; color:#c659ab; padding-top:12px; clear:both;}
...
 .qa-related-q-list {list-style-type:none; padding:0;}
-.qa-related-q-item {margin:0.5em 0; word-wrap: break-word;}
+.qa-related-q-item {
+	display: flex;
+	padding: 5px 0;
+	word-wrap: break-word;
+	border-bottom: 1px solid #e0e0e0;
+}
+.qa-related-q-item-stats {
+    flex-shrink: 0;
```
pupi1985, line 463: Here and in many places below CSS spacing is not a single tab

rxchun:
Thanks for the pointers @pupi1985 , I will patch these up.
Weird that in some places are not single tabbed. It's showing correctly on my IDE :\

PR #1002, opened 2024-07-05
Added comments page in the profile

csemasthan:
As of now, only all questions and all answers of a user can be accessed through profile section.

Now, I have added all comments section in the profile section.


qa-include/pages/user-comments.php
```diff
@@ -0,0 +1,104 @@
...
+
+foreach ($questions as $question) {
+	$options = qa_post_html_options($question, $htmldefaults);
+	$options['voteview'] = qa_get_vote_view('C', false, false);
+
+	$qa_content['q_list']['qs'][] = qa_other_to_q_html_fields($question, $loginuserid, qa_cookie_get(),
+		$usershtml, null, $options);
+}
+
+$qa_content['page_links'] = qa_html_page_links(qa_request(), $start, $pagesize, $count, qa_opt('pages_prev_next'));
```
csemasthan: $count is not correct here. replace with "count($questions)"

PR #1015, opened 2025-12-08
Fix anonymous author microdata for Google Rich Results

rxchun:
## Summary
Adds an empty `<meta itemprop="url" content="" />` for anonymous post authors to satisfy Google structured data requirements, without affecting registered users.

## Details
- When `$microdata` is true and the post author is anonymous, the generated HTML now wraps the author name in:

```
<span itemprop="author" itemscope itemtype="https://schema.org/Person">
  <span itemprop="name">anonymous</span>
  <meta itemprop="url" content="" />
</span>
```
- This prevents Google Rich Results warnings about missing `url`.
- **Registered users** (with `<a>` links) are unaffected.
- The patch does **not** change any IP display or other functionality.

### Testing
- Verified anonymous posts render the proper `<meta itemprop="url">`.
- Verified registered users continue to render their existing linked name correctly.

### Issue addressed
Google structured data requires an `author/url` field. Anonymous authors previously triggered warnings in Rich Results testing tools.


rxchun:
The URL actually needs to contain content, as reported by the webmaster who found this issue.

Since there isn't a default fallback user to reference anonymous content, or attribute stray content like StackOverflow does with the user ***[@community](https://stackoverflow.com/users/-1/community)***, the best solution here is to set the link to the **Users** page. This ensures that SEO has a valid page to reference without leading to a 404 error.

Using `mysite.com/users/anonymous` could be problematic since it might either:

 - Point to a real user with the username "anonymous," or
 - Result in a 404 error if such a user doesn’t exist.

By pointing to the general **Users** page (`mysite.com/users/`), we avoid these issues and maintain valid structured data without exposing sensitive information or creating misleading URLs.

